PR #7, opened 2024-11-27
Report correct error at startSession

gmrchk:
Some browsers won't allow calling getSupportedConstraints before getting approval for microphone access. In such case we fail with random error of getSupportedConstraints being undefined, while we should fail on access.


hikmet-demir (approved):
LGTM, thanks for the addition, I will add this to documentation as well.

PR #8, opened 2024-12-03
iOS headphones hack

gmrchk:
In majority of the cases, the system/browser selects correct input/output based on connected devices. However, iOS seems to struggle with it in browser, and instead of using connected bluetooth headphones uses iPhone mic/speaker. Annoyingly, this is not a consistent issue, and even iPhone can be forced to use Airpods at times, seemingly with no specific steps to reproduce consistently. 

This PR ads a hack into the SDK to detect iOS, and find audio devices containing "airpod" or "headphone" or "earphone" in the label, and sets those as ideal if found. Selecting a correct audio input device will switch to correct (same) audio output also. This is only applicable on iOS, but not in other browser, where i/o could be set to different devices using this method. That's why the hack is limited purely to iOS. 

I didn't want to include the hack in the SDK itself, but it's becoming too complex to follow and I suspect this will be a frequent request. 

Alternative is to provide a workaround instructions + way to set preferred device id (eg. user would have to figure that out on their own) when starting a call.

### DIFF
--- a/packages/client/README.md
+++ b/packages/client/README.md
@@ -159,6 +159,16 @@ const conversation = await Conversation.startSession({
 });
 ```
 
+#### Prefer Headphones for iOS Devices
+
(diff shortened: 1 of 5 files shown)


jodik (approved):
Thanks Georgy

aarthificial (approved):

PR #18, opened 2025-01-03
feat: Added interrupt option to Conversation and useConversation

aleksailosey:
This is useful for a "click to interrupt" feature

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -422,6 +422,10 @@ export class Conversation {
     this.lastFeedbackEventId = this.currentEventId;
     this.updateCanSendFeedback();
   };
+
+  public interrupt = () => {
+    return this.fadeOutAudio()
+  };
 }
 
 export function postOverallFeedback(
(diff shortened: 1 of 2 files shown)


imaxisXD:
I think this is added https://github.com/elevenlabs/packages/pull/35/files

imaxisXD:
Plus one this is a much needed thing in the sdk

AngeloGiacco (approved):

AngeloGiacco:
lgtm but can you give send off as well @aarthificial

packages/client/src/index.ts
```diff
@@ -438,6 +438,10 @@ export class Conversation {
     this.lastFeedbackEventId = this.currentEventId;
     this.updateCanSendFeedback();
   };
+
+  public interrupt = () => {
+    return this.fadeOutAudio()
```
aarthificial, line 443: We should set the `lastInterruptionTimestamp` to the `currentEventId` to prevent future audio chunks streamed for the current agent response from canceling the fadeout

aarthificial (changes requested):

PR #22, opened 2025-01-15
Implement automatic reconnects for the webscoket connection

CodingFu:
The problem with browsers is that internet connection is not always great and the websockets disconnect sometimes.

This PR implements automatic reconnects to broken websocket connection

### DIFF
--- a/packages/client/src/utils/connection.ts
+++ b/packages/client/src/utils/connection.ts
@@ -81,6 +81,35 @@ export class Connection {
         protocols.push(`bearer.${config.authorization}`);
       }
       socket = new WebSocket(url, protocols);
+
+      socket.addEventListener("close", event => {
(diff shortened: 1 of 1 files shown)


oscar-horizonsarchitecture:
Hello, we need this implementation ASAP

louisjoecodes:
Thank you for the PR.

Last week we deployed a fix to reduce the number of unexpected WebSocket interruptions. There should be a reduced number of issues here now across both Flash & Turbo.

We can't just add a patch to re-connect the WebSocket every time as this would restart the conversation. We'd much rather fix this at source. Thanks!
Louis

Othman2001:
@louisjoecodes We still face the problem, in almost every 10 conversations it happened twice.

PR #24, opened 2025-01-22
Improve disconnection handling

aarthificial:
Makes sure the react SDK behaves correctly when the server terminates the connection.

`onDisconnect` now receives a disconnection reason (`user`, `agent`, `error`) and a human-readable message in case of an error.

Our backend will soon start sending much clearer and more precise error messages.


gmrchk (approved):
Ohlala, looking great! 

Should we bump the version in same PR and publish?

PR #28, opened 2025-01-24
Queue incoming events

aarthificial:
- Requests the mic access first, before initiating a connection
- Queues any websocket events that may happen during AudioContext initialization

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -7,11 +7,7 @@ import {
   OnDisconnectCallback,
   SessionConfig,
 } from "./utils/connection";
-import {
-  ClientToolCallEvent,
-  isValidSocketEvent,
-  PingEvent,
-} from "./utils/events";
+import { ClientToolCallEvent, IncomingSocketEvent } from "./utils/events";
 
 export type { IncomingSocketEvent } from "./utils/events";
 export type { SessionConfig, DisconnectionDetails } from "./utils/connection";
@@ -84,6 +80,13 @@ export class Conversation {
     let output: Output | null = null;
 
     try {
+      // some browsers won't allow calling getSupportedConstraints or enumerateDevices
+      // before getting approval for microphone access
+      const preliminaryInputStream = await navigator.mediaDevices.getUserMedia({
+        audio: true,
+      });
+      preliminaryInputStream?.getTracks().forEach(track => track.stop());
+
       connection = await Connection.create(options);
       [input, output] = await Promise.all([
         Input.create({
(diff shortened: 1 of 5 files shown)


packages/client/src/index.ts
```diff
@@ -328,7 +314,7 @@ export class Conversation {
     }
   };
 
-  private addAudioBase64Chunk = async (chunk: string) => {
+  private addAudioBase64Chunk = (chunk: string) => {
```
gmrchk, line 317: Why were these turned async?

packages/client/src/utils/connection.ts
```diff
@@ -174,11 +176,13 @@ export class Connection {
...
   }
 
+  private queue: IncomingSocketEvent[] = [];
   private disconnectionDetails: DisconnectionDetails | null = null;
   private onDisconnectCallback: OnDisconnectCallback | null = null;
+  private onMessageCallback: OnMessageCallback | null = null;
 
   private constructor(
-    public readonly socket: WebSocket,
+    private readonly socket: WebSocket,
```
gmrchk, line 185: Turning it private might be (type) breaking change.

gmrchk (approved):

packages/client/src/index.ts
aarthificial: They used to be async before concatenation was done by an audio worklet and I forgot to remove it back then.

packages/client/src/utils/connection.ts
aarthificial: True, I reverted it

packages/client/src/index.ts
```diff
@@ -84,6 +80,13 @@ export class Conversation {
     let output: Output | null = null;
 
     try {
+      // some browsers won't allow calling getSupportedConstraints or enumerateDevices
+      // before getting approval for microphone access
+      const preliminaryInputStream = await navigator.mediaDevices.getUserMedia({
+        audio: true,
+      });
+      preliminaryInputStream?.getTracks().forEach(track => track.stop());
+
```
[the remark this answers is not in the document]
aarthificial: Yes, this is how it used to be before. But the input can now have a different sampling rate depending on the agent configuration, so we need to establish a connection first, before we know how to set up the streams.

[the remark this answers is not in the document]
gmrchk: Gotcha! Perhaps it could be within `Input.prepare` method or something, which would be called here? The code (at least with the current comment about getSupportedConstraints or enumerateDevices) doesn't make much sense here.

PR #31, opened 2025-02-04
Add custom delay

aarthificial:
Add a way to fix an issue on Android that cause the first few seconds of the agent message to be cut off

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -93,16 +93,15 @@ export class Conversation {
       });
 
       const delayConfig = options.connectionDelay ?? {
-        other: 0,
+        default: 0,
         // Give the Android AudioManager enough time to switch to the correct audio mode
         android: 3_000,
-        ios: 0,
       };
-      let delay = delayConfig.other;
+      let delay = delayConfig.default;
       if (isAndroidDevice()) {
-        delay = delayConfig.android;
+        delay = delayConfig.android ?? delay;
       } else if (isIosDevice()) {
-        delay = delayConfig.ios;
+        delay = delayConfig.ios ?? delay;
       }
 
       if (delay > 0) {
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -11,11 +11,7 @@ import { ClientToolCallEvent, IncomingSocketEvent } from "./utils/events";
 import { isAndroidDevice, isIosDevice } from "./utils/compatibility";
 
 export type { IncomingSocketEvent } from "./utils/events";
-export type {
-  SessionConfig,
-  DisconnectionDetails,
-  Language,
-} from "./utils/connection";
+export type { SessionConfig, DisconnectionDetails } from "./utils/connection";
 export type Role = "user" | "ai";
 export type Mode = "speaking" | "listening";
 export type Status =
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -8,9 +8,14 @@ import {
   SessionConfig,
 } from "./utils/connection";
 import { ClientToolCallEvent, IncomingSocketEvent } from "./utils/events";
+import { isAndroidDevice, isIosDevice } from "./utils/compatibility";
 
 export type { IncomingSocketEvent } from "./utils/events";
-export type { SessionConfig, DisconnectionDetails } from "./utils/connection";
+export type {
+  SessionConfig,
+  DisconnectionDetails,
+  Language,
+} from "./utils/connection";
 export type Role = "user" | "ai";
 export type Mode = "speaking" | "listening";
 export type Status =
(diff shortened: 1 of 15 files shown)


packages/client/src/index.ts
```diff
@@ -96,9 +118,13 @@ export class Conversation {
         Output.create(connection.outputFormat),
       ]);
 
+      preliminaryInputStream?.getTracks().forEach(track => track.stop());
+      preliminaryInputStream = null;
+
```
aarthificial, line 123: I moved the disposal of the `preliminaryInputStream` here in order to keep Android in call mode while we await the delay. Otherwise we'd need to await it *after* the connection has been established. Which is okay for shorter delays, but for longer ones the agent would mistakenly detect that the caller has been silent for too long and would start talking again.

packages/client/src/index.ts
```diff
@@ -78,14 +83,31 @@ export class Conversation {
...
-      const preliminaryInputStream = await navigator.mediaDevices.getUserMedia({
+      preliminaryInputStream = await navigator.mediaDevices.getUserMedia({
         audio: true,
       });
-      preliminaryInputStream?.getTracks().forEach(track => track.stop());
+
+      const delayConfig = options.connectionDelay ?? {
+        other: 0,
+        // Give the Android AudioManager enough time to switch to the correct audio mode
+        android: 3_000,
```
gmrchk, line 98: 3s seems a bit much. Is that based on some testing or documentation?

packages/client/src/index.ts
```diff
@@ -78,14 +83,31 @@ export class Conversation {
...
+      } else if (isIosDevice()) {
+        delay = delayConfig.ios;
+      }
+
+      if (delay > 0) {
+        await new Promise(resolve => setTimeout(resolve, delay));
+      }
 
       connection = await Connection.create(options);
       [input, output] = await Promise.all([
```
gmrchk, line 113: The actual stream is in the input down here. Will this still avoid the problem even when we're waiting for the `preliminaryInputStream`, which is not used and destroyed?

packages/client/src/utils/connection.ts
```diff
@@ -55,6 +55,11 @@ export type SessionConfig = {
   };
   customLlmExtraBody?: any;
   dynamicVariables?: Record<string, string | number | boolean>;
+  connectionDelay?: {
+    other: number;
+    android: number;
+    ios: number;
+  };
```
gmrchk, line 62: I'd probably expect there to be an option to set all of them at once, but not important and would likely cause rest of the code to be more complicated. Just an idea...

```suggestion
  connectionDelay?: {
    default: number;
    android?: number;
    ios?: number;
  };
```

packages/client/src/index.ts
```diff
@@ -8,9 +8,14 @@ import {
...
 } from "./utils/connection";
 import { ClientToolCallEvent, IncomingSocketEvent } from "./utils/events";
+import { isAndroidDevice, isIosDevice } from "./utils/compatibility";
 
 export type { IncomingSocketEvent } from "./utils/events";
-export type { SessionConfig, DisconnectionDetails } from "./utils/connection";
+export type {
+  SessionConfig,
+  DisconnectionDetails,
+  Language,
```
gmrchk, line 17: Language import seems unrelated.

packages/react/src/index.ts
```diff
@@ -15,6 +15,7 @@ export type {
   Status,
   SessionConfig,
   DisconnectionDetails,
+  Language,
```
gmrchk, line 18: Same here.

gmrchk (approved):
I really hoped it doesn't come to this. 😄

packages/client/src/index.ts
aarthificial: Yes, I [moved the disposal](https://github.com/elevenlabs/packages/pull/31#discussion_r1940300692) to after the actual input is created which will keep Android in call mode the whole time

packages/client/src/utils/connection.ts
aarthificial: Good point, added

packages/client/src/index.ts
aarthificial: Yup, added it because it was missing making it difficult to override the `language` in a type-safe way
Will move it to a separate PR though

gmrchk: Fine here, just wasn't sure if you're aware.

packages/client/src/index.ts
aarthificial: The initial report mentioned the first 1-3 seconds being cut off. I tested a one second delay and it sometimes worked, but sometimes cut off the first word. Maybe we could settle on 2 seconds, but I'm afraid that low-end devices may take more time than the pixel I tested it on.

gmrchk: Got it, makes sense! I guess better safe than sorry!

PR #35, opened 2025-03-03
Microphone mute/unmute

aarthificial:


packages/client/src/index.ts
```diff
@@ -320,7 +321,7 @@ export class Conversation {
     // check if the sound was loud enough, so we don't send unnecessary chunks
     // then forward audio to websocket
     //if (maxVolume > 0.001) {
-    if (this.status === "connected") {
+    if (!this.micMuted && this.status === "connected") {
       this.connection.sendMessage({
         user_audio_chunk: arrayBufferToBase64(rawAudioPcmData.buffer),
         //sample_rate: this.inputAudioContext?.inputSampleRate || this.inputSampleRate,
```
gmrchk, line 327: I remember introducing a logic in the past with `maxVolume` (above), where we wouldn't send audio if it was too quite, just as it seemed useless. However, there was some issue on the server kind of expecting the stream to be coming in. Pretty sure it shouldn't fail or anything, but I'd suggest to double check it doesn't break the conversation flow in some way.

packages/react/README.md
```diff
@@ -206,6 +206,24 @@ A method to set the output volume of the conversation. Accepts object with volum
...
+const { micMuted, setMicMuted } = useConversation();
+
+// Mute the microphone
+muteMic(true);
+
+// Unmute the microphone
+muteMic(false);
+
+// Check if the microphone is currently muted
+console.log(isMicMuted);
```
gmrchk, line 224: Example seems wrong.

gmrchk (approved):

packages/client/src/index.ts
aarthificial: Yes, the backend is not ready for this, I updated the logic to continue sending silent audio while muted

PR #41, opened 2025-03-12
Always cast tool result to string

PaulAsjes:
The API expects the result of a client tool to always be a string, this enforces that on the SDK level.

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -239,10 +239,13 @@ export class Conversation {
               ](parsedEvent.client_tool_call.parameters)) ??
               "Client tool execution successful."; // default client-tool call response
 
+            // The API expects result to be a string, so we need to convert it if it's not already a string
+            const formattedResult = typeof result === 'object' ? JSON.stringify(result) : String(result);
+
             this.connection.sendMessage({
               type: "client_tool_result",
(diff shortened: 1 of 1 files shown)


packages/client/src/index.ts
```diff
@@ -239,10 +239,13 @@ export class Conversation {
               ](parsedEvent.client_tool_call.parameters)) ??
               "Client tool execution successful."; // default client-tool call response
 
+            // The API expects result to be a string, so we need to convert it if it's not already a string
+            const formattedResult = typeof result === 'object' ? JSON.stringify(result) : String(result);
```
KacperWalentynowicz, line 243: _Minor_

Would be good to console.warn in case the result is not a string.

KacperWalentynowicz, line 243: _Question_

Will this work for arrays?

KacperWalentynowicz (approved):

packages/client/src/index.ts
PaulAsjes: As in warn before we do the conversion? I don't think that's necessary as developers should be able to return what they want. Ensuring that the result is the right type is up to us on the SDK side.

KacperWalentynowicz: ah, sorry. You're right, they don't need to provide strings I forgot about it

PR #53, opened 2025-03-26
Try to acquire a Wake Lock during conversation

aarthificial:
Also adds prettier to CI and adds missing `pt-br` language type

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -252,10 +252,6 @@ export class Conversation {
       }
 
       case "client_tool_call": {
-        console.info(
-          "Received client tool call request",
-          parsedEvent.client_tool_call
-        );
         if (
           this.options.clientTools.hasOwnProperty(
             parsedEvent.client_tool_call.tool_name
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -12,7 +12,11 @@ import { isAndroidDevice, isIosDevice } from "./utils/compatibility";
 
 export type { InputConfig } from "./utils/input";
 export type { IncomingSocketEvent } from "./utils/events";
-export type { SessionConfig, DisconnectionDetails, Language } from "./utils/connection";
+export type {
+  SessionConfig,
+  DisconnectionDetails,
+  Language,
+} from "./utils/connection";
 export type Role = "user" | "ai";
 export type Mode = "speaking" | "listening";
 export type Status =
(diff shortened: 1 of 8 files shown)


KacperWalentynowicz (commented):
I understand the idea but would feel more comfortable if someone who knows wake locks also reviewed.

packages/client/src/index.ts
```diff
@@ -229,7 +250,10 @@ export class Conversation {
       }
 
       case "client_tool_call": {
-        console.info("Received client tool call request", parsedEvent.client_tool_call);
+        console.info(
+          "Received client tool call request",
+          parsedEvent.client_tool_call
+        );
```
gmrchk, line 256: Actually don't think this should be here, should it?

packages/react/package.json
```diff
@@ -19,10 +19,10 @@
     "build": "BROWSERSLIST_ENV=modern microbundle --jsx React.createElement --jsxFragment React.Fragment --jsxImportSource react src/index.ts",
     "clean": "rm -rf ./dist",
     "dev": "npm run clean && BROWSERSLIST_ENV=development microbundle --jsx React.createElement --jsxFragment React.Fragment --jsxImportSource react src/index.ts -w -f modern",
-    "lint": "npm run lint:ts && npm run lint:es",
+    "lint": "npm run lint:ts && npm run lint:es && npm run lint:prettier",
     "lint:ts": "tsc --noEmit --skipLibCheck",
     "lint:es": "npx eslint .",
-    "lint:prettier": "prettier 'src/**/*.ts'",
+    "lint:prettier": "prettier 'src/**/*.ts' --check",
```
gmrchk, line 25: 👍

packages/client/src/index.ts
```diff
@@ -119,13 +132,16 @@ export class Conversation {
...
-      return new Conversation(fullOptions, connection, input, output);
+      return new Conversation(fullOptions, connection, input, output, wakeLock);
     } catch (error) {
       fullOptions.onStatusChange({ status: "disconnected" });
       preliminaryInputStream?.getTracks().forEach(track => track.stop());
       connection?.close();
       await input?.close();
       await output?.close();
+      try {
+        await wakeLock?.release();
```
gmrchk, line 143: FYI docs also mention setting the wakeLock reference to null.

gmrchk (approved):
That's pretty cool. Had no idea that's thing. 

For the record, was there actually some issues with this being a problem or why are we adding it?

packages/client/src/index.ts
aarthificial: Good point

packages/client/src/index.ts
aarthificial: Removed

aarthificial:
@gmrchk we got some complaints about the connection terminating when the client's screen locks due to inactivity, despite them talking. Not much we can do about the browser throttling and terminating the ws connection but keeping the screen awake should help a bit

PR #55, opened 2025-04-01
Interface to set audio input/output device

ashish-kumar-eightfold:
Adds method in `Conversation` to change audio input/output device.


jake-luster:
This is a crucial feature for any real-world voice application. Handling mid-call device switching is a common user expectation, and the lack of support for updating the audio stream dynamically is a significant limitation of the current SDK.

We’d really appreciate it if this PR could be reviewed and merged — it would unlock a much smoother experience for many teams, including ours, who see this as a core missing capability.

PR #58, opened 2025-04-11
Add fallback for iframes in Safari

aarthificial:
Attempting to start a conversation in Safari inside an iframe will throw a CORS error because the `blob://` protocol is considered cross-origin.

In such cases, we fall back to using a base64 data URL which is not as widely supported as the blob but works in Safari.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@11labs/client",
-  "version": "0.1.2",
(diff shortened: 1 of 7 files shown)


gmrchk (approved):

KacperWalentynowicz:
Looks good, feel free to rely on Georgy's expertise here and merge, my review wouldn't add much :)

PR #60, opened 2025-04-16
feat: add onInterruption callback

enijar:
Adds `onInterruption` callback to the API.


louisjoecodes:
This is implemented thank you!

PR #62, opened 2025-04-23
Add widget scaffolding

aarthificial:


packages/convai-widget-embed/package.json
```diff
@@ -0,0 +1,42 @@
+{
+  "name": "@11labs/convai-widget-embed",
+  "version": "0.0.1",
```
gmrchk, line 3: Just a thought - might wanna start from 1 here to avoid confusion? The users of the widget are a bit less technical and talking about version 1 seems more logical than version 0 through support etc.

gmrchk (approved):
LGTM, but didn't test properly (can still do so, approving for the sake of time).

Do I get it right there will be another `convai-widget-react` package, and so on, for specific framework usage?

packages/client/src/utils/connection.ts
```diff
@@ -37,6 +37,8 @@ export type Language =
   | "uk"
   | "ru"
   | "hu"
+  | "hr"
+  | "sk"
```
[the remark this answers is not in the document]
aarthificial: The languages ConvAI supports are an intersection of what Whisper supports and what our TTS models support. Not sure why these two were missing here but with them this type covers all 32 languages we support

packages/convai-widget-embed/package.json
aarthificial: Agree but I wanna keep it at 0.0.1 while testing. The plan is to start developing the widget here and first use it in-house before doing a proper release and switching our clients to it. For the time being, the recommending way to embed is still the old widget

aarthificial:
>Do I get it right there will be another convai-widget-react package, and so on, for specific framework usage?

Yes! I'd like to do something similar to [NumberFlow](https://number-flow.barvian.me) where framework-specific packages provide a wrapper for the underlying web component

PR #64, opened 2025-04-29
Implement non-expandable variants of the widget

aarthificial:
All contexts are taken as-is from our current internal implementation.

TODO in future PRs:
- [ ] Finish up expanded mode, starting with transcriptions only and then extending to chat mode when ready.
- [ ] Update the Orb shader to match our internal one and support custom images
- [ ] Add feedback collection - waiting for designs
- [ ] Add error states - waiting for designs


kamilk-11:
@nevflynn Yes, sorry, I need to update the video, it's present in the current version:
Screenshot 2025-05-19 at 14 48 09

packages/convai-widget-core/src/types/attributes.ts
```diff
@@ -26,18 +34,12 @@ const MainAttributeList = [
   "override-first-message",
   "override-language",
   "override-voice-id",
+  "mic-muting",
+  "transcript",
+  "text-input",
+  "text-contents",
 ] as const;
```
gmrchk, line 41: Are these (or can be) attributes `data-` prefixed? If not, it might cause some HTML validators to mark them as invalid. If that's the case, some customer might complain that they are chasing perfect SEO scores or whatnot, and that this shows up for them.

packages/convai-widget-core/src/types/config.ts
```diff
@@ -40,12 +60,34 @@ export type AvatarConfig =
...
+  copied: "Copied!",
+  accept_terms: "Accept",
+  dismiss_terms: "Cancel",
+
+  listening_status: "Listening",
+  speaking_status: "Talk to interrupt",
+  connecting_status: "Connecting",
+
+  input_label: "Text message input",
+  input_placeholder: "Or send a message",
```
gmrchk, line 80: I find the "Or" as the start of the placeholder to look weird in the UI. Did we consider some other options?

packages/convai-widget-core/src/types/config.ts
```diff
@@ -40,12 +60,34 @@ export type AvatarConfig =
     };
 
 export const DefaultTextContents = {
```
gmrchk, line 62: Wonder how are these going to be translated based on selected language. Are they?

packages/convai-widget-core/src/widget/LanguageSelect.tsx
```diff
@@ -0,0 +1,70 @@
...
+  const portal = useRootPortal();
+
+  return (
+    <Select.Root
+      open={open}
+      value={language.value.languageCode}
+      onValueChange={value => {
+        setLanguage(value as Language);
+        setOpen(false);
+      }}
```
gmrchk, line 27: Unsure if you tried to go through the flow with screen readers, but since it's going to be a complex form (chat/voice interface) around the web, might be useful. Some blind users would might end up no able to contact support.

packages/convai-widget-core/src/widget/SheetActions.tsx
```diff
@@ -0,0 +1,77 @@
...
+  } = useConversation();
+
+  return (
+    <div className="shrink-0 overflow-hidden flex p-3 items-end justify-end">
+      {text_input_enabled && (
+        <TextArea
+          rows={1}
+          aria-label={text.input_label}
+          value={userMessage}
+          onInput={sendUserActivity}
```
gmrchk, line 38: Don't remember at the moment, will paste work well with onInput?

packages/convai-widget-core/src/widget/Transcript.tsx
```diff
@@ -0,0 +1,49 @@
...
+interface TranscriptProps {
+  scrollPinned: Signal<boolean>;
+  transcript: ReadonlySignal<TranscriptEntry[]>;
+}
+
+export function Transcript({ scrollPinned, transcript }: TranscriptProps) {
+  const scrollContainer = useRef<HTMLDivElement>(null);
+  const scrollToBottom = () => {
+    scrollContainer.current?.scrollTo(0, scrollContainer.current.scrollHeight);
+  };
```
gmrchk, line 15: I haven't tried yet, so have no idea how it looks, but I feel like these scroll jumps/transitions influence a lot how the whole interaction feels. Would be great to polish the timings, especially if messages are sliding in or something.

packages/convai-widget-core/tailwind.config.js
```diff
@@ -1,3 +1,57 @@
...
+  "subtle",
+  "primary",
+  "error",
+];
+const spacing = [
+  0, 0.5, 1, 1.5, 2, 2.5, 3, 3.5, 4, 5, 6, 7, 8, 9, 10, 11, 12, 14, 16, 20, 24,
+  28, 32, 36, 40, 44, 48, 52, 56, 60, 64, 72, 80, 96,
+];
+const radius = [1, 2, 2.5, 3, 4, 5, 6, 7, 7.5];
+
```
gmrchk, line 16: I've hit some issues in the past where global styles of the parent page would be set to some non-standard value, and it would be carried out to shadow too, unless reset within. 
I don't remember the detail, but we could test and possibly introduce our won reset.

packages/convai-widget-core/src/components/Flag.tsx
```diff
@@ -0,0 +1,32 @@
...
+  ...props
+}: FlagProps) {
+  return (
+    <img
+      className={clsx(
+        "rounded-full object-cover",
+        SIZE_CLASSES[size],
+        className
+      )}
+      src={`https://purecatamphetamine.github.io/country-flag-icons/1x1/${flagCode.toUpperCase()}.svg`}
```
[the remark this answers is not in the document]
gmrchk: There is some library we use in the codebase directly (in conv AI lang selectors for example). Just FYI, in case we want to perhaps unify. 

I think we could just grab these file from the library and upload them to our own CDN. Should be really straight forward, and I can help if you happen to not have access or something.

But you're right, all looks legit and used according to docs! Just the guys profile didn't really promote confidence, and since technically the images could be replaced with something innapropriate for example, I'd hate to get into situation where we have to explain to all our enterprise customers why do they have some porn images on their website or something.

packages/convai-widget-core/tailwind.config.js
gmrchk: Ah yeah rem must be the thing I remember. Seems like you got that covered!

Note - I would not expect any issues on docs or xi - those use pretty much the same setup with tailwind, etc.

PR #73, opened 2025-05-22
Add text-only mode

kamilk-11:
Adds support for text-only mode in the SDK together with a corresponding chat mode in the widget.

TODO:
- [x] Update README

### DIFF
--- a/packages/client/src/VoiceConversation.ts
+++ b/packages/client/src/VoiceConversation.ts
@@ -102,7 +102,6 @@ export class VoiceConversation extends BaseConversation {
   }
 
   protected override handleAudio(event: AgentAudioEvent) {
-    // await super.handleAudio(event);
     if (this.lastInterruptTimestamp <= event.audio_event.event_id) {
       this.options.onAudio(event.audio_event.audio_base_64);
       this.addAudioBase64Chunk(event.audio_event.audio_base_64);
--- a/packages/convai-widget-core/src/widget/Transcript.tsx
+++ b/packages/convai-widget-core/src/widget/Transcript.tsx
@@ -3,6 +3,8 @@ import { TranscriptEntry } from "../contexts/conversation";
 import { useEffect, useRef } from "preact/compat";
 import { TranscriptMessage } from "./TranscriptMessage";
 
+const SCROLL_PIN_PADDING = 16;
+
 interface TranscriptProps {
   scrollPinned: Signal<boolean>;
   transcript: ReadonlySignal<TranscriptEntry[]>;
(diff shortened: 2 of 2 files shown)


packages/client/src/VoiceConversation.ts
```diff
@@ -0,0 +1,202 @@
...
+    await this.output.close();
+  }
+
+  protected override handleInterruption(event: InterruptionEvent) {
+    super.handleInterruption(event);
+    this.fadeOutAudio();
+  }
+
+  protected override handleAudio(event: AgentAudioEvent) {
+    // await super.handleAudio(event);
```
gmrchk, line 105: Is this important?

packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -29,19 +30,39 @@ export type TranscriptEntry =
...
+      conversationIndex: number;
     };
 
 export function ConversationProvider({ children }: ConversationProviderProps) {
   const value = useConversationSetup();
+
+  // Automatically disconnect the conversation after 10 minutes of no messages
+  useSignalEffect(() => {
+    if (value.conversationTextOnly.value === true) {
+      value.transcript.value;
```
gmrchk, line 53: What does this line do? 🤔

packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -182,7 +253,14 @@ function useConversationSetup() {
...
           error.value = message;
-          transcript.value = [...transcript.value, { type: "error", message }];
+          transcript.value = [
+            ...transcript.value,
+            {
+              type: "error",
+              message,
+              conversationIndex: conversationIndex.peek(),
+            },
+          ];
```
gmrchk, line 263: Wonder if we can run into some rendering issues when the array is updated. React is not best with those, so might be issue with longer conversations, especially if we want to animate things. 

Updating messages (ex. loading state if something is loading for long time, and eventually maybe fails to send) may be tricky too. Perhaps some message IDs is something to consider?

packages/convai-widget-core/src/contexts/session-config.tsx
```diff
@@ -35,6 +38,9 @@ export function SessionConfigProvider({
     tts: {
       voiceId: overrideVoiceId.value,
     },
+    conversation: {
+      textOnly: parseBoolAttribute(overrideTextOnly.value) ?? undefined,
```
gmrchk, line 42: Double-checking - this will be overridden if user start the convo by text even if set to false, right?

gmrchk (approved):
Did not test, approving to unblock and release for testing!

PR #107, opened 2025-07-07
Add user-id to conv-ai sdks

jameszhou02:


packages/client/README.md
```diff
@@ -209,6 +209,18 @@ const conversation = await Conversation.startSession({
 });
 ```
 
+#### User identification
```
PaulAsjes, line 212: I think we need to explain where you'd get said user ID and what happens if you do provide the ID. We have this [endpoint](https://elevenlabs.io/docs/api-reference/user/get) but that doesn't explain how to get the ID itself.

PaulAsjes (changes requested):
One nit on docs. Once this is merged I can merge [this PR](https://github.com/elevenlabs/packages/pull/105) and release as 0.3.0.

PaulAsjes (changes requested):

packages/client/README.md
```diff
@@ -209,6 +209,20 @@ const conversation = await Conversation.startSession({
...
+
+You can optionally pass a user ID to identify the user in the conversation. This can be your own customer identifier. This will be included in the conversation initiation data sent to the server:
+
+Tracking this id can be helpful for filtering conversations, tracking analytics on a user level, etc.
```
PaulAsjes, line 216:
```suggestion
Tracking this ID can be helpful for filtering conversations, tracking analytics on a user level, etc.
```

PaulAsjes (approved):

PR #118, opened 2025-07-21
[convai] fix token params

AngeloGiacco:


packages/client/src/utils/WebRTCConnection.ts
```diff
@@ -58,9 +58,9 @@ export class WebRTCConnection extends BaseConnection {
     } else if ("agentId" in config && config.agentId) {
       // Agent ID provided - fetch token from API
       try {
-        const response = await fetch(
-          `https://api.elevenlabs.io/v1/convai/conversation/token?agent_id=${config.agentId}`
-        );
+        const version = process.env.npm_package_version || "0.3.0";
+        const url = `https://api.elevenlabs.io/v1/convai/conversation/token?agent_id=${config.agentId}&source=react_sdk&version=${version}`;
```
PaulAsjes, line 62: Why `react_sdk`? You can use packages/client without using React.

packages/client/src/utils/WebRTCConnection.ts
```diff
@@ -58,9 +58,9 @@ export class WebRTCConnection extends BaseConnection {
     } else if ("agentId" in config && config.agentId) {
       // Agent ID provided - fetch token from API
       try {
-        const response = await fetch(
-          `https://api.elevenlabs.io/v1/convai/conversation/token?agent_id=${config.agentId}`
-        );
+        const version = process.env.npm_package_version || "0.3.0";
```
PaulAsjes, line 61: This will fail as `npm_package_version` only works in npm script executions. Instead you should look into build-time injection via package.json.

Also I think you should default to null or empty string instead of hardcoding the version here.

PaulAsjes (changes requested):

packages/convai-widget-core/src/version.ts
```diff
@@ -0,0 +1,2 @@
+// This file is auto-generated during build
+export const PACKAGE_VERSION = "0.0.13";
```
PaulAsjes, line 2: Should update this.

packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@elevenlabs/react",
-  "version": "0.3.0",
+  "version": "0.3.1",
```
PaulAsjes, line 3: This adds functionality, so should be 0.4.0

PaulAsjes (changes requested):

packages/client/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@elevenlabs/client",
-  "version": "0.3.0",
+  "version": "0.3.1",
```
PaulAsjes, line 3: React and client SDKs should be in lock-step, so this should be 0.4.0 as well.

PR #123, opened 2025-07-24
expo react native demo nits

thorwebdev:
Just some small nits based on how I usually work with expo.

### DIFF
--- a/examples/react-native-expo/App.tsx
+++ b/examples/react-native-expo/App.tsx
@@ -2,7 +2,6 @@ import React, { useState } from "react";
 import { View, Text, StyleSheet, TouchableOpacity, Alert } from "react-native";
 import { TextInput } from "react-native";
 import { ElevenLabsProvider, useConversation } from "@elevenlabs/react-native";
-import { Platform } from "react-native";
 import type {
   ConversationStatus,
   Callbacks,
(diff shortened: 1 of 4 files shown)


examples/react-native-expo/App.tsx
```diff
@@ -1,47 +1,66 @@
...
     if (isStarting) return;
 
     setIsStarting(true);
     try {
       await conversation.startSession({
-        agentId: '<your-agent-id>', // Replace with your actual agent ID
+        agentId: process.env.EXPO_PUBLIC_AGENT_ID,
+        connectionType: "webrtc",
+        dynamicVariables: {
+          platform: Platform.OS,
```
PaulAsjes, line 59: This will only work if you've set up the agent to have dynamic variables right? Think we should remove this for the example app.

PaulAsjes (approved):
One nit, otherwise good!

examples/react-native-expo/App.tsx
thorwebdev: The reason I like to keep it in, is that it shows folks what's possible. It works, even if you don't have dynamic variables set. Wdyt?

PaulAsjes: Think this is a good thing to include in the cookbook where we can add context.

PR #128, opened 2025-07-25
Fix overrides

PaulAsjes:


jodik (approved):

packages/react-native/src/utils/overrides.ts
```diff
@@ -21,7 +21,7 @@ export function constructOverrides(
       conversation: {
         text_only: config.overrides.conversation?.textOnly,
       },
-      client: {
+      source_info: {
         source: config.overrides?.client?.source || "react_native_sdk",
```
AngeloGiacco, line 25:
```suggestion
      source_info: {
source: "react_native_sdk",
```

if we get here should just always be react_native_sdk

AngeloGiacco (approved):

PR #148, opened 2025-07-31
add residency options to react

AngeloGiacco:
adds residency option and fixes onConnect

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@elevenlabs/react",
-  "version": "0.4.3",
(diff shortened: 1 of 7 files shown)


packages/react/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@elevenlabs/react",
-  "version": "0.4.3",
```
AngeloGiacco, line 3: not sure why this is showing 0.4.3, it is currently 0.4.4 https://github.com/elevenlabs/packages/blob/main/packages/react/package.json

ISNIT0 (approved):

PR #150, opened 2025-07-31
[React Native] make webrtc endpoints configurable

ISNIT0:


AngeloGiacco:
looks good but will only dispatch agent to on of our livekit servers, so should maybe add a note that it is not quite as permissive as this PR makes it look

PR #171, opened 2025-08-15
[convai widget] optionally use rtc if public

AngeloGiacco:
## Summary
- Adds optional WebRTC support controlled by `use_rtc` config field
- WebSocket remains default, WebRTC enabled when `use_rtc: true`
- Maintains backward compatibility

## Changes
- Added `use_rtc` to WidgetConfig interface
- Created useWebRTC hook
- Updated SessionConfigProvider for conditional connection type
- SignedUrl connections remain WebSocket-only

### DIFF
--- a/packages/convai-widget-core/package.json
+++ b/packages/convai-widget-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@elevenlabs/convai-widget-core",
-  "version": "0.1.3",
+  "version": "0.1.4",
   "description": "The common library for the Conversational AI Widget.",
   "main": "./dist/index.js",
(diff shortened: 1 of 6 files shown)


packages/convai-widget-core/package.json
```diff
@@ -1,6 +1,6 @@
 {
   "name": "@elevenlabs/convai-widget-core",
-  "version": "0.1.3",
+  "version": "0.1.4",
```
PaulAsjes, line 3: This change is a minor, not a patch.

PaulAsjes (approved):

Issue #115, opened 2025-07-19
`setVolume()` does not work.


jhobz23:
Hi everyone,

Just a heads-up: `conversation.setVolume()` doesn’t seem to work. Here’s the snippet I’m using — the volume slider change is handled by `handleVolumeChange()`:

```
const [volume, setVolumeState] = useState(0.5);
const conversation = useConversation({ volume }); // initial volume works

const handleVolumeChange = (value: number[]) => {
  conversation.setVolume({ volume });  // no effect during an active session
  setVolumeState(value[0] / 100);
  console.log("Volume set to:", volume);
};
```
Also, I noticed in the [npm documentation](https://www.npmjs.com/package/@elevenlabs/react?activeTab=readme) it shows:

```
const [volume, setVolume] = useState(0.5);
const conversation = useConversation({ volume });

// Set the volume
setVolume(0.5);
```
But `setVolume(0.5)` only updates the local state — it doesn’t actually change the conversation volume during an active session.

Am I missing something, or is the documentation possibly out of sync?

I'm using version 0.3.0 by the way.

Thanks!

PaulAsjes:
Hi there, can you show how you're initiating the conversation?

jhobz23:
> Hi there, can you show how you're initiating the conversation?

Sorry for the late reply, here's the snippet:

```
const [volume, setVolumeState] = useState(0.5);
const conversation = useConversation({ volume });

const handlePlay = async () => {
        try {

        const response = await fetch(route('call-sim.signed-url',aiAgent.uuid));
        const signedUrl = await response.text();

        if (!response.ok) {
            toast.error("Conversation failed to start.",{description:"Error fetching signed url, please try again later."});
        }
            setLoading(true);
            const id = await conversation.startSession({
                signedUrl: signedUrl,
                userId: auth.user.uuid,
                connectionType: 'websocket',
                onDisconnect: () => {
                    setConversationId(id);
                    toast.info("Conversation disconnected.",{description: "The conversation has been disconnected."});
                    console.log("Conversation ID:", id);
                },
                onMessage: (message) => {
                      setMessages((prevMessages) => [
                        ...prevMessages,
                        {
                        id: Date.now(),
                        sender: message.source === 'ai' ? 'AI Agent' : 'User',
                        content: message.message,
                        created_at: new Date().toLocaleTimeString(),
                        user_name: 'You',
                        ai_agent_name: aiAgent.name,
                        },
                    ]);
                },
            });
[...]

Issue #109, opened 2025-07-09
Client side tools called inconsistently


doriansmiley:
# Summary
I am experiencing an issue where the client-side tools defined in my agent are called inconsistently. Sometimes the model executes them just fine, while others it never calls them at all. Below is my agent setup
```json
{
  "agent_id": "<Ommited>",
  "name": "<Ommited>",
  "conversation_config": {
    "asr": {
      "quality": "high",
      "provider": "elevenlabs",
      "user_input_audio_format": "pcm_16000",
      "keywords": []
    },
    "turn": {
      "turn_timeout": 30,
      "silence_end_call_timeout": -1,
      "mode": "turn"
    },
    "tts": {
      "model_id": "eleven_flash_v2",
      "voice_id": "<Ommited>",
      "supported_voices": [],
      "agent_output_audio_format": "pcm_16000",
      "optimize_streaming_latency": 3,
      "stability": 0.5,
      "speed": 1,
      "similarity_boost": 0.8,
      "pronunciation_dictionary_locators": []
    },
    "conversation": {
      "text_only": false,
      "max_duration_seconds": 3600,
      "client_events": [
        "audio",
        "interruption",
        "user_transcript",
        "agent_response",
        "agent_response_correction"
      ]
    },
    "language_presets": {},
    "agent": {
      "first_message": "Hey there {{userName}}. How can I help you today?",
      "language": "en",
      "dynamic_variables": {
        "dynamic_variable_placeholders": {
          "localDay": "Monday",
          "userId": "<Ommited>",
          "userName": "<Ommited>",
          "locatDateTime": "4/15/2025, 10:14:06 AM"
        }
      },
      "prompt": {
        "prompt": "<Ommited>",
        "llm": "gemini-2.0-flash-001",
        "temperature": 0.5,
        "max_tokens": -1,
        "tool_ids": [
          "<Ommited>",
          "<Ommited>"
        ],
        "built_in_tools": {
          "end_call": {
            "name": "end_call",
            "description": "",
            "response_timeout_secs": 20,
            "type": "system",
            "params": {
              "system_tool_type": "end_call"
            }
          },
          "language_detection": null,
          "transfer_to_agent": null,
          "transfer_to_number": null,
          "skip_turn": null,
          "play_keypad_touch_tone": null
        },
        "mcp_server_ids": [],
        "native_mcp_server_ids": [],
        "knowledge_base": [],
        "custom_llm": null,
        "ignore_default_personality": false,
        "rag": {
[...]

PR #184, opened 2025-08-28
Fix setVolume and add client tool debugging

PaulAsjes:
Fixes https://github.com/elevenlabs/packages/issues/115

Attempts to fix https://github.com/elevenlabs/packages/issues/109 by adding more debug logic to determine what the problem is.

### DIFF
--- a/packages/client/src/VoiceConversation.ts
+++ b/packages/client/src/VoiceConversation.ts
@@ -222,4 +222,16 @@ export class VoiceConversation extends BaseConversation {
   public getOutputVolume() {
     return this.calculateVolume(this.getOutputByteFrequencyData());
   }
+
+  public setVolume = ({ volume }: { volume: number }) => {
+    this.volume = volume;
+
+    if (this.connection instanceof WebRTCConnection) {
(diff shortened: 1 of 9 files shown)


packages/client/src/VoiceConversation.ts
```diff
@@ -222,4 +222,16 @@ export class VoiceConversation extends BaseConversation {
   public getOutputVolume() {
     return this.calculateVolume(this.getOutputByteFrequencyData());
   }
+
+  public setVolume = ({ volume }: { volume: number }) => {
```
louisjoecodes, line 226: Would try to clamp & coerce the input volume as it can take anything at the moment:
```
// VoiceConversation.ts
public setVolume = ({ volume }: { volume: number }) => {
  // clamp & coerce
  const v = Number.isFinite(volume) ? Math.min(1, Math.max(0, volume)) : 1;
  this.volume = v;

  if (this.connection instanceof WebRTCConnection) {
    this.connection.setAudioVolume(v);
  } else if (this.output?.gain?.gain) {
    this.output.gain.gain.value = v;
  }
};
```

Issue #186, opened 2025-08-31
Setting output device


s-devbay:
I’m attempting to select a specific audio device when starting the session, but the audio always plays through the default device. Below is the code snippet I’m using. Is this functionality supported, or am I missing something?

```
await navigator.mediaDevices.getUserMedia({
     audio: {
            deviceId: { exact: this.selectedMicId },
            echoCancellation: true,
            noiseSuppression: true,
            autoGainControl: true
     }
});

await Conversation.startSession({
     signedUrl: signedUrl,
     outputDeviceId: this.selectedSpeakerId,
     textOnly: false,
});
```

PaulAsjes:
Added this in 0.6.1, you can now provide the `outputDeviceId` in `startSession` like you have in your example.

Issue #110, opened 2025-07-09
Audio data mismatch between @11labs/client and @elevenlabs/client packages


enijar:
I've just moved from the deprecated package [@11labs/client](https://www.npmjs.com/package/@11labs/client) [v0.1.4] to the successor package [@elevenlabs/client](https://www.npmjs.com/package/@elevenlabs/client) [v0.2.2].

There is an issue with calls to the [getOutputByteFrequencyData](https://github.com/elevenlabs/packages/blob/8864b3c071722b480952a8f38d7e7173d550525b/packages/client/src/VoiceConversation.ts#L191) method returning a `Uint8Array` with no data (zeros)

The previous [@11labs/client](https://www.npmjs.com/package/@11labs/client) package provided access to the underlying `output` property, which gave access to the `AudioContext` API, which correctly got the output frequency like this:

```typescript
import { Conversation } from "@11labs/client";
const conversation = await Conversation.startSession({
  // ...
});
conversation.output.analyser.fftSize = 1024;
const bufferLength = conversation.output.analyser.frequencyBinCount;
const dataArray = new Uint8Array(bufferLength);
(function tick() {
  conversation.output.analyser.getByteTimeDomainData(dataArray);
  console.log(dataArray); // Uint8Array
  requestAnimationFrame(tick);
})();
```

Can the [getOutputByteFrequencyData](https://github.com/elevenlabs/packages/blob/8864b3c071722b480952a8f38d7e7173d550525b/packages/client/src/VoiceConversation.ts#L191) method be fixed so it returns non-zero output data?

acostajohn:
Hey guys, any timeline for this? We'd like to move to WebRTC as soon as possible which is only supported by the latest SDKs!

PR #187, opened 2025-09-01
Fix output bytes and device input/output switching

PaulAsjes:
Fixes https://github.com/elevenlabs/packages/issues/110
Fixes https://github.com/elevenlabs/packages/issues/186

Makes the `getOutputByteFrequencyData` methods work in WebRTC mode. Also fixes setting input/output devices.

### DIFF
--- a/packages/client/src/utils/input.ts
+++ b/packages/client/src/utils/input.ts
@@ -25,6 +25,9 @@ export class Input {
         sampleRate: { ideal: sampleRate },
         echoCancellation: true,
         noiseSuppression: true,
+        autoGainControl: true,
+        // Enhanced constraints for better echo cancellation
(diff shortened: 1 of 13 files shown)


packages/client/src/utils/input.ts
```diff
@@ -86,15 +90,60 @@ export class Input {
...
+
+    try {
+      // Create new constraints with the specified device
+      const options: MediaTrackConstraints = {
+        deviceId: { exact: inputDeviceId },
+        echoCancellation: true,
+        noiseSuppression: true,
+        autoGainControl: true,
+        channelCount: { ideal: 1 },
+      };
```
AngeloGiacco, line 122: seems duplicated, maybe can unify?

AngeloGiacco (approved):
lgtm left a couple of comments

Issue #193, opened 2025-09-04
blob: and data: require too relaxed Content-Security-Policy


julesvirallinen:
Hi! Working on an integration, and running into a bit of a hiccup. The scripts are sent to the worklets with a blob (fallbacking to a data-url). Both of these are risky to enable in content security policies, and can't be whitelisted for specific domains at all (all or nothing). 

From https://www.w3.org/TR/CSP2/#source-list-guid-matching

> 4.2.2.1. Security Considerations for GUID URL schemes
>This section is not normative.
>As defined above, special URL schemes that refer to specific pieces of unique content, such as "data:", "blob:" and "filesystem:" are excluded from matching a policy of * and must be explicitly listed. Policy authors should note that the content of such URLs is often derived from a response body or execution in a Document context, which may be unsafe. Especially for the [default-src](https://www.w3.org/TR/CSP2/#default_src) and [script-src](https://www.w3.org/TR/CSP2/#script_src) directives, policy authors should be aware that allowing "data:" URLs is equivalent to unsafe-inline and allowing "blob:" or "filesystem:" URLs is equivalent to unsafe-eval.

To address this, we'd need another fallback to the existing ones that doesn't pose the same risks. I'm working on a fork that allows saving the files to `public/worklets` and checks them as a fallback. It works locally, but does that sound like a viable direction, or are there other suggestions on how to address this?

PaulAsjes:
Sorry for the delay! Just reviewed.

PaulAsjes:
Fix released in 0.7.0!

PR #195, opened 2025-09-05
Allow passing script paths to avoid blob: and data:

julesvirallinen:
relates to #193

Allows passing worklets paths via options to client, allowing for secure CSPs.


packages/client/worklets/raw-audio-processor.js
```diff
@@ -0,0 +1,123 @@
+/*
```
PaulAsjes, line 1: 1. Can we keep to the same naming scheme and use camel case for the file name?
2. I'm not keen that we now have duplicated code here and in `utils/rawAudioProcessor.ts`. Would be great if we could unify the two instead. Same goes for the `audioConcatProcessor`.

PR #208, opened 2025-09-15
[agents cli] renaming

AngeloGiacco:


AngeloGiacco:
idk what happened here there should be a huge diff

PR #213, opened 2025-09-15
Map client SDK types to newly generated events

louisjoecodes:
- Maps existing types to newly generated SDK types


PaulAsjes (approved):
So much cleaner!

PR #216, opened 2025-09-15
chore: add missing callbacks to client

louisjoecodes:


PaulAsjes (approved):
Will need to add these to the React package as well.

PR #217, opened 2025-09-15
[agents cli] convai cli -> agents cli 

AngeloGiacco:
outline of PR, may still need another iteration after some more thorough testing
only non-ui diff is renaming convai.lock -> agents.lock but should be non-breaking as its just a lock file


PaulAsjes (changes requested):
Make sure you update the root readme as well.

PR #219, opened 2025-09-15
chore: rm unused node-wav dependency

louisjoecodes:


louisjoecodes:
Turns out we might need it...

PR #225, opened 2025-09-16
[Widget] make widget dismissible

louisjoecodes:
This is opt-in, a 'dismissible' parameter must be added to the widget for this behaviour to take effect:
CleanShot 2025-09-16 at 09 55 17@2x


PaulAsjes:
@kamilk-11 Here's a good example of why we need to collapse the widget: https://github.com/elevenlabs/elevenlabs-docs/issues/1356

Also some Discord comments:
CleanShot 2025-09-17 at 14 45 13

CleanShot 2025-09-17 at 14 45 33

Agree on the front that the widget should be uncollapsible.

PR #243, opened 2025-09-24
[Widget] Add expand event

PaulAsjes:
Adds a new `elevenlabs-agent:expand` event to trigger the widget expanding without clicking the "Start call" button. Only works if the widget is able to expand, i.e. if text-input, text-only or transcript modes are enabled.

### DIFF
--- a/packages/convai-widget-core/src/index.dev.tsx
+++ b/packages/convai-widget-core/src/index.dev.tsx
@@ -25,6 +25,7 @@ function Playground() {
   const [textInput, setTextInput] = useState(false);
   const [textOnly, setTextOnly] = useState(false);
   const [alwaysExpanded, setAlwaysExpanded] = useState(false);
+  const [expanded, setExpanded] = useState(false);
 
   return (
     <div className="w-screen h-screen flex items-center justify-center bg-base-hover text-base-primary">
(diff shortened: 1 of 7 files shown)


packages/react-native/src/components/LiveKitRoomWrapper.tsx
```diff
@@ -57,7 +57,7 @@ export const LiveKitRoomWrapper = ({
         clientTools={clientTools}
         updateCurrentEventId={updateCurrentEventId}
       />
-      {children}
+      {children as any}
```
AngeloGiacco, line 60: is this necessary?

packages/convai-widget-core/src/index.dev.tsx
```diff
@@ -25,6 +25,7 @@ function Playground() {
   const [textInput, setTextInput] = useState(false);
   const [textOnly, setTextOnly] = useState(false);
   const [alwaysExpanded, setAlwaysExpanded] = useState(false);
+  const [expanded, setExpanded] = useState(false);
```
AngeloGiacco, line 28: can we merge these two?

AngeloGiacco (approved):
looks good to me

PR #252, opened 2025-09-29
feat: add components cli for ui.elevenlabs.io

louisjoecodes:
- Implements: `agents components add <component>`

### DIFF
--- a/packages/agents-cli/src/cli.ts
+++ b/packages/agents-cli/src/cli.ts
@@ -2446,7 +2446,7 @@ if (args.length === 0 || args.includes('--help') || args.includes('-h')) {
   const componentsCommand = program
     .command("components")
     .description(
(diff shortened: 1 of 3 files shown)


packages/agents-cli/src/cli.ts
```diff
@@ -2440,5 +2442,63 @@ if (args.length === 0 || args.includes('--help') || args.includes('-h')) {
     console.error('  client-tool    Add a new client tool');
     process.exit(1);
   }
+
+  const componentsCommand = program
+    .command("components")
+    .description(
+      "Import components from the ElevenLabs UI (https://ui.elevenlabs.io)"
```
PaulAsjes, line 2449:
```suggestion
      "Import components from the ElevenLabs UI registry (https://ui.elevenlabs.io)"
```

PR #310, opened 2025-10-15
fix: handle max_duration_seconds timeout on client side (#298)

### LINKED ISSUE #298: Call continues showing as "connected" on client after max_duration_seconds is reached, despite server marking it as "Successful"
When a WebRTC call reaches the max_duration_seconds limit configured in the agent, the server correctly ends the call and marks it as "Successful" in the ElevenLabs dashboard. However, the client never receives any disconnect event and the conversation remains in "connected" state indefinitely.

Steps to reproduce:
1. Configure agent with max_duration_seconds: 3600 (or any duration)
2. Start a WebRTC session using Conversation.startSession()
3. Keep the call active until the max duration is reached
4. Server ends the call automatically
5. Check ElevenLabs dashboard - shows call as "Successful" with correct duration (e.g., 60:00)
6. Check client state - still shows as "connected", no callbacks were triggered

Expected behavior:
When max_duration_seconds is reached and the server ends the call, the client should receive:
1. onDisconnect() callback, OR
2. onStatusChange() callback with status "disconnected", OR
3. onError() callback with information about max duration reached

Actual behavior:
1. No callbacks are triggered
2. Client remains in "connected" state
3. No events indicating the call has ended
4. ElevenLabs dashboard correctly shows the call as completed with "Successful" status

Environment:
Package: @elevenlabs/client@0.7.1
Connection type: WebRTC
Agent config: max_duration_seconds: 3600

diko0071:
- Add ErrorMessage event handling to client
- Client now receives disconnect event when max_duration_seconds is reached
- ErrorMessageEvent added to IncomingSocketEvent union
- max_duration_exceeded error triggers onDisconnect callback instead of leaving client in connected state

Fixes #298

### DIFF
--- a/packages/client/src/BaseConversation.ts
+++ b/packages/client/src/BaseConversation.ts
@@ -299,13 +299,14 @@ export class BaseConversation {
 
   protected handleErrorEvent(event: ErrorMessageEvent) {
     const errorType = event.error_event.error_type;
-    const message = event.error_event.message || event.error_event.reason || "Unknown error";
-    
+    const message =
+      event.error_event.message || event.error_event.reason || "Unknown error";
+
     if (errorType === "max_duration_exceeded") {
       this.endSessionWithDetails({
-        reason: "agent",
+        reason: "error",
         message: message,
-        context: event,
+        context: new Event("max_duration_exceeded"),
       });
       return;
     }


PaulAsjes (approved):
Thanks very much!

diko0071:
@PaulAsjes sry, was issue with types. Now all should be good, works locally!

Issue #227, opened 2025-09-16
Normalization Breaks Standard HTTP Headers


oksanamar3718241:
https://github.com/elevenlabs/packages/blob/0be9d6714e33b14204e7324f4dc549baa517474f/packages/convai-cli/src/utils.ts#L284-L308

The toCamelCaseKey() function removes hyphens and converts keys to camelCase. When this is applied to request_headers, it changes standard HTTP headers like Content-Type or X-Api-Key into ContentType and xApiKey, which stops them from working correctly.

PR #316, opened 2025-10-17
[agents cli] fix headers camelcased

PennyroyalTea:
headers should not be camelcased, as reported by one of the customer. fix + cover with unit test


PennyroyalTea:
https://github.com/elevenlabs/packages/issues/227

AngeloGiacco (approved):

PR #321, opened 2025-10-21
Fix post tool call sound speech fade out

vijaypemmaraju:
the js sdk currently has an issue where an interruption event triggers a fade out which ramps the audio down and back up over 2 seconds, which can disrupt agent speech. to mitigate this, i have introduced an interruption type [in xi](https://github.com/elevenlabs/xi/pull/19632) to differentiate between a user interruption and a tool call sound interruption, this pr utilizes said type

### DIFF
--- a/packages/client/src/VoiceConversation.ts
+++ b/packages/client/src/VoiceConversation.ts
@@ -122,7 +122,16 @@ export class VoiceConversation extends BaseConversation {
 
   protected override handleInterruption(event: InterruptionEvent) {
     super.handleInterruption(event);
-    this.fadeOutAudio();
+
+    const interruptionType =
+      event.interruption_event?.interruption_type ?? "user";
+
+    if (interruptionType === "tool_sound") {
+      this.output.worklet.port.postMessage({ type: "interrupt" });
+      this.output.worklet.port.postMessage({ type: "clearInterrupted" });
+    } else {
+      this.fadeOutAudio();
+    }
   }
 
   protected override handleAudio(event: AgentAudioEvent) {
--- a/packages/types/generated/types/asyncapi-types.ts
+++ b/packages/types/generated/types/asyncapi-types.ts
@@ -214,7 +214,10 @@ export interface Interruption {
 
 export interface InterruptionEvent {
   event_id: number;
+  interruption_type?: InterruptionEventInterruptionType;
 }
+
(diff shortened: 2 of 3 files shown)


packages/client/src/VoiceConversation.ts
```diff
@@ -122,7 +119,18 @@ export class VoiceConversation extends BaseConversation {
 
   protected override handleInterruption(event: InterruptionEvent) {
     super.handleInterruption(event);
-    this.fadeOutAudio();
+
+    const interruptionId = event.interruption_event?.event_id;
+    const isToolSoundInterrupt =
+      typeof interruptionId === "number" &&
+      interruptionId < this.currentEventId;
```
sedatcagdas: I don't think it's a good strategy. we can't really say `interruptionId < this.currentEventId;` is always tool interruption. do we know why we receive/process interruption event after the post tool speech?

vijaypemmaraju: this _is_ the interruption event to stop the tool sound. 

and the two lines below will [clear the buffer](https://github.com/elevenlabs/packages/blob/053363e4686bbe8da0f401d06fcc7ef6e87595ff/packages/client/src/utils/audioConcatProcessor.ts#L57-L58)

packages/client/src/VoiceConversation.ts
```diff
@@ -122,7 +122,16 @@ export class VoiceConversation extends BaseConversation {
 
   protected override handleInterruption(event: InterruptionEvent) {
     super.handleInterruption(event);
-    this.fadeOutAudio();
+
+    const interruptionType =
+      event.interruption_event?.interruption_type ?? "user";
```
vijaypemmaraju: always default to `user` for backwards compatibility

packages/client/src/VoiceConversation.ts
```diff
@@ -122,7 +122,16 @@ export class VoiceConversation extends BaseConversation {
...
+
+    if (interruptionType === "tool_sound") {
+      this.output.worklet.port.postMessage({ type: "interrupt" });
+      this.output.worklet.port.postMessage({ type: "clearInterrupted" });
```
PaulAsjes: Is the idea that in case of tool sound interrupt we stop the audio immediately instead of fading out?

PR #331, opened 2025-10-31
[Agents] CSAT Form after convo completes

kalvin807:
Adds CSAT feedback section.
- blocked by https://github.com/elevenlabs/xi/pull/20089

- slightly changes the style of avatar header per new design.
- changed postOverallFeedback to send feedback with rating and type
- adds CSAT feedback section.
  - a 1-5 rating component with star icon as default
- add a fullscreen followup page for user to enter extra comments

### DIFF
--- a/packages/convai-widget-core/src/widget/FeedbackPage.tsx
+++ b/packages/convai-widget-core/src/widget/FeedbackPage.tsx
@@ -40,7 +40,7 @@ export function FeedbackPage() {
           </div>
         </div>
         <TextArea
-          className="w-full min-h-[2lh]"
+          className="w-full min-h-[6lh]"
           placeholder={text.follow_up_feedback_placeholder}
           rows={6}
           value={feedbackText}


packages/convai-widget-core/src/widget/TranscriptMessage.tsx
```diff
@@ -46,33 +50,50 @@ export function TranscriptMessage({
...
-            <span className="break-all">
-              {text.conversation_id}: {lastId.value}
-            </span>
+      );
+    }
+
+    if (entry.type === "disconnection") {
+      return (
+        <div className="mt-2 px-8 flex flex-col">
+          {entry.role !== "user" && shouldShowFeedbackAtEnd.value && (
```
kamilk-11: Question, since I'm not up-to-date with requirements, we only want to show the feedback if it was the agent who ended the call?

packages/convai-widget-core/src/widget/Sheet.tsx
```diff
@@ -91,52 +96,63 @@ export function Sheet({ open }: SheetProps) {
...
+            <Transcript
+              transcript={filteredTranscript}
+              scrollPinned={scrollPinned}
+            />
+            <SheetActions
+              scrollPinned={scrollPinned}
+              showTranscript={showTranscript}
+            />
+          </>
+        )}
```
kamilk-11: Let's use `InOutTransition` to fade this in out:
```suggestion
        <InOutTransition active={currentContent.value === "transcript"}>
          <div className="grow flex flex-col min-h-0 transition-opacity duration-300 ease-out data-hidden:opacity-0">
            <Transcript
              transcript={filteredTranscript}
              scrollPinned={scrollPinned}
            />
            <SheetActions
              scrollPinned={scrollPinned}
              showTranscript={showTranscript}
            />
          </div>
        </InOutTransition>
```

packages/convai-widget-core/src/components/Rating.tsx
```diff
@@ -0,0 +1,179 @@
...
+      role="radio"
+      aria-checked={isFilled}
+      aria-label={`${value} star${value !== 1 ? "s" : ""}`}
+      tabIndex={0}
+      onMouseEnter={() => onHover(value)}
+      onClick={handleClick}
+      onKeyDown={handleKeyDown}
+      className={clsx(
+        "w-8 h-8 grid place-content-center text-base-primary transition-opacity",
+        !isFilled && !isHovered && "opacity-15"
```
kamilk-11: Two things, let's add `rounded-full and focus-ring` to improve replace the default blue focus outline with the same style as other components use.

Also I'd suggest using `text-base-border` instead of `opacity-15` to not affect the visibility of the focus style:

```suggestion
        "w-8 h-8 grid place-content-center rounded-full focus-ring text-base-primary transition-opacity",
        !isFilled && !isHovered && "text-base-border"
```

packages/convai-widget-core/src/contexts/feedback.tsx
```diff
@@ -0,0 +1,123 @@
...
+  return (
+    <FeedbackContext.Provider
+      value={{
+        rating,
+        feedbackText,
+        feedbackProgress,
+        submitRating,
+        submitFeedback,
+        reset,
+      }}
```
kamilk-11: Same thing with this provider, there's nothing here that requires a variable reference, you could define everything in a `useMemo` and never cause re-renders

kamilk-11 (commented):
Solid work!

packages/convai-widget-core/src/widget/FeedbackPage.tsx
```diff
@@ -0,0 +1,57 @@
...
+            <p className="text-sm text-base-primary font-medium">
+              {text.thanks_for_feedback}
+            </p>
+            <p className="text-sm text-base-subtle">
+              {text.thanks_for_feedback_details}
+            </p>
+          </div>
+        </div>
+        <TextArea
+          className="w-full min-h-[6lh]"
```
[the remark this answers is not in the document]
kalvin807: CleanShot 2025-11-06 at 01 16 31@2x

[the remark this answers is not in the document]
kamilk-11: I see, the way I interpreted the design was that the box would take up the space up until the button even when empty:
Screenshot 2025-11-07 at 08 26 25

But growing on user text is okay too. I'd revert back the previous height because some browsers don't support `field-sizing:content` so they'd still get some more space for feedback

PR #339, opened 2025-11-05
[Agents] Widget: Added overrides handling for text customisation attributes

alexgoodison:
Added the appropriate override handling for various text customisation attributes available for the Agents widget.


kalvin807 (commented):
LGTM. 


Personally prefer 1to1 mapping with internal labels, but documentation already defined the naming at https://elevenlabs.io/docs/agents-platform/customization/widget#core-configuration

```
main_label > main-label-text
start_call > start-call-text
end_call > end-call-text
expand > expand-text
listening_status > listening-status-text
speaking_status > speaking-status-text
```

kalvin807:
cc @kamilk-11 for an extra look

packages/convai-widget-core/src/contexts/text-contents.tsx
```diff
@@ -37,12 +44,26 @@ export function TextContentsProvider({ children }: TextContentsProviderProps) {
...
 
+  const attributeTextOverrides = useComputed(() => {
+    const overrides: Partial<TextContents> = {};
+
+    if (actionText.value) overrides.main_label = actionText.value;
+    if (startCallText.value) overrides.start_call = startCallText.value;
+    if (endCallText.value) overrides.end_call = endCallText.value;
+    if (expandText.value) overrides.expand = expandText.value;
+    if (listeningText.value) overrides.listening_status = listeningText.value;
+    if (speakingText.value) overrides.speaking_status = speakingText.value;
```
kamilk-11: What's the context for this feature?

There are many more text attributes than those six:
https://github.com/elevenlabs/packages/blob/dd8a28ea44ef6ad059ba8782cbb141dda5939969/packages/convai-widget-core/src/types/config.ts#L81-L119

And they are already overrideable via the `text-contents` attribute. Defining specific overrides for those six seems a bit confusing

kamilk-11: Agree, that's an example from the legacy widget

PR #351, opened 2025-11-13
fix(convai-widget-core): ignore empty stream messages

mrvissercb:
For reasons I have yet to nail down, the elevenlabs websocket can deliver empty streamed messages (i.e., a "start" followed by a "stop" immediately after). It may just be the LLM deciding to send a "text" step with an empty string, and just some funny wording embedded somewhere in our prompting... 

One additional thing about it is while there is both a start/stop onAgentResponsePart message, there is no followup onMessage that triggers -- so this may have been a thing that existed before, but the back-end filters out empty onMessage events?

I can continue trial and error there to see why the LLM wants to give us an empty message, but in the meantime it probably makes sense to not allow empty text messages to stream into the transcript? That's what this fix does.

### DIFF
--- a/packages/convai-widget-core/src/index.dev.tsx
+++ b/packages/convai-widget-core/src/index.dev.tsx
@@ -27,6 +27,8 @@ function Playground() {
   const [alwaysExpanded, setAlwaysExpanded] = useState(false);
   const [dynamicVariablesStr, setDynamicVariablesStr] = useState("")
   const [expanded, setExpanded] = useState(false);
+  const [overrideFirstMessage, setOverrideFirstMessage] = useState(false);
+  const [firstMessage, setFirstMessage] = useState("Hi, how can I help you today?")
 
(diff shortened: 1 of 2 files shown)


packages/convai-widget-core/src/index.dev.tsx
```diff
@@ -113,6 +115,27 @@ function Playground() {
...
+                <input
+                  type="text"
+                  className="p-1 bg-base border border-base-border"
+                  value={firstMessage}
+                  disabled={!overrideFirstMessage}
+                  onChange={e => setFirstMessage(e.currentTarget.value)}
+                />
+              )}
+            </div>
+        </label>
```
mrvissercb: I had added this to try and reproduce using our same configuration with a first-message, but was unable to. Figured I'd leave the override into the dev UI. Can remove or improve if you like.

PR #352, opened 2025-11-13
fix(convai-widget-core): respect feedback mode to determine if feedback form should be displayed

mrvissercb:
The handling of feedback_mode doesn't seem to be inline with the widget config API. On my agent, end_feedback is always `null`, but `feedback_mode` will be set to none/during/end.

I assume the API changed, so I've done my best to kind of guess what the right logic should be if one were using the older API. Let me know if I got it wrong.

Screen recording (sorry embedded vids don't work for me): [Screen Recording 2025-11-13 at 1.52.18 PM.mov.gz](https://github.com/user-attachments/files/23533049/Screen.Recording.2025-11-13.at.1.52.18.PM.mov.gz)


kalvin807 (commented):
Thank you for the PR! the changes overall looks good and will fix asap

PR #353, opened 2025-11-13
fix(convai-widget-core): auto-scroll on transcript usually stops scrolling

mrvissercb:
The `onScroll` on the transcript scroll container is triggering for both automated and manual scrolls.

It looks like the `isAtBottom` heuristic can become `false` via the automated means of the transcript container size changing and the smart scroll falling behind.

onWheel is only triggered on manual user events, but admittedly isn't all-encompassing. This is simple and is better than where it was, though.

An alternative I thought of would be to keep a ref of the previous scroll Y, and then cancel automatic scroll if we detect it going up. Happy to try that out if we really don't like onWheel.

Before: 
[Screen Recording 2025-11-13 at 2.30.08 PM.zip](https://github.com/user-attachments/files/23533527/Screen.Recording.2025-11-13.at.2.30.08.PM.zip)

After:
[Screen Recording 2025-11-13 at 2.43.30 PM.zip](https://github.com/user-attachments/files/23533530/Screen.Recording.2025-11-13.at.2.43.30.PM.zip)


PaulAsjes (approved):
Thanks! I had to double check that this works for trackpads as well.

Issue #341, opened 2025-11-05
React Native Architecture Issue: Provider Contains Too Much State, Causing Unnecessary Rerenders


coryetzkorn:
### Problem
`ElevenLabsProvider` contains 8+ pieces of frequently-changing state, causing all children to rerender on every state change. This is a fundamental architectural issue that can't be fully fixed with memoization alone. I tried to fix it in #340 but it's still not a full fix.

Unlike other React Native libraries (see PostHog, React Query, Redux), ElevenLabsProvider puts call state inside the provider component, which causes performance issues.

### Current Architecture (Problematic)
```typescript
export const ElevenLabsProvider = ({ children }) => {
  // ❌ These state changes cause provider to rerender
  const [token, setToken] = useState('');
  const [connect, setConnect] = useState(false);
  const [status, setStatus] = useState('disconnected');      // Changes during call
  const [conversationId, setConversationId] = useState('');
  const [isSpeaking, setIsSpeaking] = useState(false);       // Toggles constantly
  const [canSendFeedback, setCanSendFeedback] = useState(false);
  const [serverUrl, setServerUrl] = useState(DEFAULT_SERVER_URL);
  const [tokenFetchUrl, setTokenFetchUrl] = useState(undefined);
  
  // When any state changes, provider rerenders → all children rerender
  return (
    <ElevenLabsContext.Provider value={contextValue}>
      <LiveKitRoomWrapper {...props}>
        {children}  {/* ← Rerenders on every state change */}
      </LiveKitRoomWrapper>
    </ElevenLabsContext.Provider>
  );
};
```

**Every `setState` call causes the provider to rerender, which causes all children to rerender by default in React.**

### Example: PostHog's Approach (Better)

PostHog's `PostHogProvider` can wrap an entire app without performance issues because it doesn't have internal state:

```typescript
// PostHog's approach (simplified)
export function PostHogProvider({ children, apiKey, options }) {
  // ✅ Creates client ONCE on mount - stable reference
  const client = useMemo(() => new PostHog(apiKey, options), [apiKey]);
  
  // ✅ Context value is stable - doesn't change during usage
  const value = useMemo(() => ({ client }), [client]);
  
  // ✅ No state = no rerenders = wrapping entire app is fine
  return (
    <PostHogContext.Provider value={value}>
      {children}
    </PostHogContext.Provider>
  );
}
```

**PostHog can wrap an entire app because the provider never rerenders.**

### Proposed Architecture (Following PostHog's Pattern)

[...]

PR #365, opened 2025-11-22
fix(react-native): stabilize provider to prevent app reload during calls

### LINKED ISSUE #362: App navigates to home screen when connecting ElevenLabs call in React Native 0.5.2
When I initiate a voice call using the @elevenlabs/react-native SDK (v0.5.2) in a bare React Native app, the call connects correctly — the microphone works, and the bot responds — but the app unexpectedly refreshes and navigates back to the home screen instead of staying on the current screen.

```
import 'react-native-get-random-values';
import 'react-native-gesture-handler';
import { AppRegistry } from 'react-native';
import App from './App';
import { name as appName } from './app.json';
import { QueryProvider, TamaguiUiProvider } from './src/shared/provider';
import { ElevenLabsProvider } from '@elevenlabs/react-native';

const AppView = () => {
  return (
    <QueryProvider>
      <ElevenLabsProvider>
        <TamaguiUiProvider>
          <App />
        </TamaguiUiProvider>
      </ElevenLabsProvider>
    </QueryProvider>
  );
};

AppRegistry.registerComponent(appName, () => AppView);
```

Expected behavior
The app should stay on the current screen while the call is active.

Observed behavior
The app refreshes and navigates to the home screen immediately after connecting the call.

React Native version: 0.79.0
SDK version: @elevenlabs/react-native@0.5.2
The call connects successfully, microphone works, bot responds — only the navigation is affected.
I suspect it might be related to the ElevenLabsProvider wrapping the app at the top level, but I am not sure how to fix it.

7ttp:
Fixes issue where ElevenLabsProvider causes app to reload and navigate to home screen when wrapping the entire application. Internal state changes during calls (isSpeaking, status, canSendFeedback) were recreating the context value on every render, forcing all consumers including React Navigation to re-render and reset.

Changes:
- Memoize contextValue with React.useMemo to prevent recreation on every render
- Memoize conversation object with proper dependencies
- Wrap inline callback functions (sendContextualUpdate, sendUserMessage, sendUserActivity) in React.useCallback for stable references
- Memoize getId function to ensure stable reference
- Move setClientTools call in useConversation to useEffect for proper dependency tracking

With these optimizations, internal state changes only trigger re-renders in components that explicitly depend on those values, preventing navigation state resets while maintaining backward compatibility.

Tests pass and no TypeScript errors.

Fixes #362
Related to #341

### DIFF
--- a/packages/react-native/src/ElevenLabsProvider.tsx
+++ b/packages/react-native/src/ElevenLabsProvider.tsx
@@ -66,10 +66,14 @@ export const useConversation = (options: ConversationOptions = {}): Conversation
     }
   }, [context, tokenFetchUrl]);
 
-  if (clientTools) {
-    context.setClientTools(clientTools);
-  }
+  React.useEffect(() => {
+    if (clientTools) {
+      context.setClientTools(clientTools);
+    }
+  }, [context, clientTools]);
 
+  // Update callbacks - since this only updates a ref, it's safe to call on every render
+  // and doesn't cause re-renders of the provider or its consumers
   context.setCallbacks(callbacks);
 
   return context.conversation;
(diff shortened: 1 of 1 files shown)


7ttp:
@PaulAsjes could you please take a quick look at this when you get a chance?

PaulAsjes (approved):

Issue #359, opened 2025-11-19
[React Native] Incorrect Status Reporting and onDisconnect Not Firing When Agent Disconnects


twalk2:
### Description
We're experiencing two critical issues with the React Native SDK's conversation state management:

1. **Status incorrectly reports as 'disconnected' when connection is active** - The `onStatusChange` callback reports `status: 'disconnected'` even though the conversation is actively connected and functioning
2. **onDisconnect callback fails to fire when the agent disconnects** - When the agent ends the conversation naturally (or due to timeout/error), the `onDisconnect` callback does not fire. **Note: `onDisconnect` DOES fire correctly when we manually call `endSession()`, but NOT when the agent disconnects on their own.**

### Environment
- **SDK**: `@elevenlabs/react-native` 
- **Platform**: React Native

### Steps to Reproduce

1. Initialize the SDK with `useConversation` hook including `onConnect`, `onDisconnect`, and `onStatusChange` callbacks
2. Start a conversation session using `startSession()`
3. Monitor the status updates via `onStatusChange`
4. **Let the agent disconnect the call naturally** (do NOT call `endSession()`)
5. Observe that `onDisconnect` never fires

### Expected Behavior

1. **Status should accurately reflect connection state**:
   - `'connecting'` → `'connected'` → `'disconnected'` 
   
2. **onDisconnect should fire when the agent ends the conversation**:
   - When agent ends naturally (finished speaking)
   - When agent times out (silence detection, inactivity)
   - When agent encounters an error
   - When we manually call `endSession()` (✅ this works)

### Actual Behavior

1. **Status is incorrectly reported**:
   - `onStatusChange` fires with `status: 'disconnected'` even while the conversation is active and working
   - Audio is flowing, messages are being exchanged, but status shows disconnected
   
2. **onDisconnect fires inconsistently**:
   - ✅ **Works**: When we call `endSession()` manually, `onDisconnect` fires correctly
   - ❌ **Broken**: When the agent ends the conversation naturally/automatically, `onDisconnect` does NOT fire
   - This makes it impossible to detect when the agent has decided to end the conversation

### Minimal Reproduction Example

Here's a complete, minimal example that demonstrates the issue:

import React, { createContext, useContext, useRef, useState, useCallback } from 'react';
import { View, Button, Text, StyleSheet } from 'react-native';
import { useConversation } from '@elevenlabs/react-native';
import type { ConversationStatus, Mode } from '@elevenlabs/react-native';

interface ConversationState {
[...]

PR #384, opened 2025-11-26
Disconnect on participant leaving

PaulAsjes:
Fixes https://github.com/elevenlabs/packages/issues/359

There was a bug in the livekit server where when the agent disconnect due to a timeout (i.e. user didn't say anything for ~30 seconds) the agent would never disconnect and the client connection would stay open. 

That's been fixed server side, this adds listeners to when a remote participant leaves. If it does, it means the agent disconnected so we call `onDisconnected` on the client and close the connection locally.

### DIFF
--- a/packages/react-native/src/components/MessageHandler.tsx
+++ b/packages/react-native/src/components/MessageHandler.tsx
@@ -50,11 +50,19 @@ export const MessageHandler = ({
   // Track agent response count for synthetic event IDs (WebRTC mode)
   const agentResponseCountRef = React.useRef(1);
 
+  // Refs for callbacks to avoid unnecessary effect re-runs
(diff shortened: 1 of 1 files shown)


examples/react-native-expo/App.tsx
```diff
@@ -51,14 +45,14 @@ const ConversationScreen = () => {
     }) => {
       console.log(`🔊 Can send feedback: ${canSendFeedback}`);
     },
-    onVadScore: ({ vadScore }: { vadScore: number }) => {
-      console.log(`🎙️ VAD Score: ${vadScore}`);
-    },
+    // onVadScore: ({ vadScore }: { vadScore: number }) => {
+    //   console.log(`🎙️ VAD Score: ${vadScore}`);
```
AngeloGiacco: how come?

PR #400, opened 2025-12-02
[Agent widget] more stable scroll lock

kalvin807:
## Why
- Scroll lock does not work in transcript mode
- Sometime it will lose pin

## What
- Rewrote the detection to use ResizeObserver (no need to track multiple item from transcript length)

### Current behavior 
- Start pinned 
  - Keep content un-scrolled for very long first message in text only mode.
- Content grows → auto-scroll to bottom
- User scrolls UP → unpins
- User scrolls back to bottom OR sends a message → re-pins


kamilk-11 (approved):
Nice! Much cleaner

PR #408, opened 2025-12-08
resume context upon suspend

kai-stra:
Continue playing audio in case browser suspends for some reason


AngeloGiacco (changes requested):
can you add a loom of this fixing the issue? or at least the behaviour with this change

kraenhansen:
This PR either needs a rebase + review or a close - @kai-stra will you take action on this?

PR #411, opened 2025-12-09
Localise the chat terms and conditions

Hashir-Majeed:
Use the language presets chat terms added in https://github.com/elevenlabs/xi/pull/22314 to localise the terms and conditions of the chat widget


alexgoodison (approved):
nice, looks good to pass to convai reviews

PR #412, opened 2025-12-10
[agents] request mic access if != text mode

AngeloGiacco:


PaulAsjes (changes requested):
Can we add a test as well please?

Issue #410, opened 2025-12-09
Issue with ElevenLabsProvider remounting children components on `startSession`


cvincentcoleman:
It appears that when calling `await conversation.startSession({...` all children components of the ElevenLabsProvider are unmounted and remounted. 

This is an issue when trying to trigger starting a session onMount (in a useEffect) of a component, because it creates an infinite loop of the startSession call being repeatedly called over and over.

I'm confident this isn't a useEffect dependency issue. Excited about this SDK, any help here would be great!

PaulAsjes:
Hi all, I think I see the root cause. Will have a fix up later today.

PaulAsjes:
This should be fixed in v0.5.6!

PR #414, opened 2025-12-11
[RN] Fix startSession infinite loop when used in a useEffect hook

PaulAsjes:
Fixes https://github.com/elevenlabs/packages/issues/410

Also downgrades react to 19.0.0. The previous fix was for a react CVE that's only exploitable when using react server components, which RN doesn't use. Upgrading caused all sorts of issues with Expo and React version imcompatibilities.


AngeloGiacco (approved):
Lgtm but would check with sec

PR #415, opened 2025-12-12
[Agent] disallow link rendering as default

kalvin807:
## Why

Widget accept a list of config to block a link rendering as default. Widget link rendering will be limited to same domain by default to avoid fraud

## What
-  take `allowed_link_domains` and configurate block via rehype-harden

### DIFF
--- a/packages/convai-widget-core/package.json
+++ b/packages/convai-widget-core/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@elevenlabs/convai-widget-core",
-  "version": "0.6.0-beta.6",
+  "version": "0.6.0-beta.7",
   "description": "The common library for the Conversational AI Widget.",
   "main": "./dist/index.js",
   "types": "./dist/index.d.ts",
--- a/packages/convai-widget-embed/package.json
+++ b/packages/convai-widget-embed/package.json
@@ -1,6 +1,6 @@
 {
   "name": "@elevenlabs/convai-widget-embed",
-  "version": "0.6.0-beta.6",
+  "version": "0.6.0-beta.7",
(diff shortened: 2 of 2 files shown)


packages/convai-widget-core/src/markdown/utils/allowedDomainsToLinkPrefixes.ts
```diff
@@ -0,0 +1,42 @@
+function maybeParseUrl(value: string): URL | null {
+  try {
+    return new URL(value);
+  } catch {
+    return null;
+  }
+}
+
+export function allowedDomainsToLinkPrefixes(
+  domains?: string[] | null
```
AngeloGiacco: the use hook returned `undefined` but in other places we are checking string or null, can we standardise on either using null or undefined

AngeloGiacco (approved):

packages/convai-widget-core/src/markdown/utils/allowedDomainsToLinkPrefixes.ts
```diff
@@ -0,0 +1,42 @@
...
+    }
+
+    const domain = trimmed.replace(/\.$/, "");
+
+    candidates.add(`https://${domain}`);
+    candidates.add(`http://${domain}`);
+    if (!domain.startsWith("www.")) {
+      candidates.add(`https://www.${domain}`);
+      candidates.add(`http://www.${domain}`);
+    }
```
kamilk-11: This feels a bit loose in terms of security, maybe we could add `include_www_domains` and `allow_http` as settings to the agent in case people want to be strict and explicit with their allowlist?

PR #417, opened 2025-12-13
Fix: end_call system tool causes last message to be omitted

Hashir-Majeed:
When the agent uses the end_call tool, the last message it sends is not sent in `agent_response_part`s and is instead sent as a single `agent_response`, causing it not to be rendered. This PR catches that case and renders the text

Loom [working]: https://www.loom.com/share/49064b7ff97b49018b37d8e293adb439

### DIFF
--- a/packages/convai-widget-core/src/mocks/browser.ts
+++ b/packages/convai-widget-core/src/mocks/browser.ts
@@ -119,7 +119,7 @@ export const Worker = setupWorker(
           agent_response_event: { agent_response: config.first_message },
         })
       );
-      if (config.text_only) {
+      if (config.text_only && agentId !== "end_call_test") {
         client.send(
           JSON.stringify({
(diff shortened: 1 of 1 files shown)


packages/convai-widget-core/src/contexts/conversation.tsx
AngeloGiacco: should be .peek() here as well then?

packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -215,6 +215,19 @@ function useConversationSetup() {
               }
```
AngeloGiacco: if you can add a test plz would be great

AngeloGiacco (approved):
can you add a check that this does not cause issues with voice mode

PR #431, opened 2025-12-28
feat: Add support for audio alignment event

bigbernnn:
Parse incoming `audio` events to return alignment data through a dedicated callback.


packages/react-native/src/types.ts
```diff
@@ -108,6 +108,12 @@ export type ConversationMetadataEvent = ConversationMetadata;
 export type AsrInitiationMetadataEvent = AsrMetadataEvent;
 export type AgentChatResponsePartEvent =
   Incoming.AgentChatResponsePartClientEvent;
+export type AudioAlignmentData = Types.AudioEventAlignment;
```
PaulAsjes: Can we be consistent and use `AudioAlignmentEvent`?

PaulAsjes (changes requested):

PR #432, opened 2025-12-28
refactor(react-complete-example): extract events and disconnect when unmount

alamenai:
## What

1. Extract the inline type of Mode.
2. Extract the event function.
3. Cleanup when unmount.


PaulAsjes (approved):
Thanks!

PR #435, opened 2026-01-05
feat: Be able to switch to voice conversation when you start a text conversation in the widget

Hashir-Majeed:
https://www.loom.com/share/414fb261faca491e832bd3a2031a7b60

**Context**

Currently, when you start a conversation in _voice_ mode, you can both talk and type. However, if your first message is a text one, then it starts as a _text_ conversation, and you can't speak for the whole conversation.

This PR adds a button that will switch the conversation to voice. Once enabled, the button will disappear, unless the conversation failed to transfer for some reason [i.e. backend error], in which case it will be visible so the user can try again.

**Approach**

The key method that facilitates all this is `upgradeFromTextConversation` [`packages/client/src/VoiceConversation.ts`] that lets a `VoiceConversation` instance be formed out of an existing `TextConversation`. This works by requesting extracting common fields, requesting mic perms, creating the correct audio io components, etc.

Notification of the backend happens by sending a `conversation_config_update` event with the `text_only: false` flag. PR: https://github.com/elevenlabs/xi/pull/23370

### DIFF
--- a/packages/client/src/BaseConversation.test.ts
+++ b/packages/client/src/BaseConversation.test.ts
@@ -1,10 +1,40 @@
 import { describe, it, expect, vi } from "vitest";
 
 import { BaseConversation, Options, PartialOptions } from "./BaseConversation";
+import type { BaseConnection } from "./utils/BaseConnection";
+import type { ConversationConfigUpdateAppliedEvent } from "./utils/events";
+
+function createMockConnection(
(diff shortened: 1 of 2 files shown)


PaulAsjes:
Hi @Hashir-Majeed, thanks for the contribution! Could you please fix the conflicts and add some tests for the upgrade flow before we continue?

PR #446, opened 2026-01-15
Wrap error from fetching conversation token

kraenhansen:
Debugging issues with my Android emulator (having issues connecting to the service), I had a hard time determining the origin of the error (since only a message is passed through the callback).

Merging this PR will:
- Wrap the fetch of the conversation token with a try-catch adding additional context to any message thrown in the block
- Increased robustness of code reading out the server-side `message` from the a `details` object.


kraenhansen:
@cursoragent could you review this?

PR #463, opened 2026-01-21
[Widget] show agent working status

kalvin807:
## Why
Show agent activity during long async tool calls to give users feedback that the system is working.

## What
Added show_agent_status config option that displays a status badge during tool execution

Tracks tool calls via onAgentToolRequest/onAgentToolResponse events, deduplicate to 1 badge per turn.

### Known Limitation
Text stream events lack event_id, so I can't stick status badge below to same-turn messages is unreliable. Will fix after backend update.

https://github.com/user-attachments/assets/41618192-a6a6-4b20-b1e4-ed9d2bb48b50

figma: https://www.figma.com/design/4OHH52XIkdWSErnAPzPXta/Conversational-AI--V2-?node-id=7779-3354&m=dev

### DIFF
--- a/packages/convai-widget-core/src/widget/TranscriptMessage.tsx
+++ b/packages/convai-widget-core/src/widget/TranscriptMessage.tsx
@@ -6,7 +6,8 @@ import { useAvatarConfig } from "../contexts/avatar-config";
 import type { TranscriptEntry } from "../contexts/conversation";
 import { ToolCallStatus, useConversation } from "../contexts/conversation";
 import { useTextContents } from "../contexts/text-contents";
-import { useMarkdownLinkConfig, useEndFeedbackType } from "../contexts/widget-config";
+import { useMarkdownLinkConfig, useEndFeedbackType, useWidgetConfig } from "../contexts/widget-config";
+import { stripAudioTags } from "../utils/stripAudioTags";
 import { WidgetStreamdown } from "../markdown";
 
 interface TranscriptMessageProps {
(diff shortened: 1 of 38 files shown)


packages/convai-widget-core/src/widget/Transcript.tsx
```diff
@@ -8,6 +8,20 @@ interface TranscriptProps {
...
+}: {
+  entry: TranscriptEntry;
+  prevEntry: TranscriptEntry;
+}) {
+  if (entry.type === "tool_call" || prevEntry.type === "tool_call") {
+    return <div className="h-2" />;
+  }
+
+  return <div className="h-6" />;
+}
```
kamilk-11: You could achieve the same thing by keeping the `gap-6` and adding the following classes to `ToolCallMessage`:
```
-my-4 first:mt-0 last:mb-0
```

packages/convai-widget-core/src/widget/TranscriptMessage.tsx
```diff
@@ -126,15 +127,55 @@ function ModeToggleMessage({ entry }: ModeToggleMessageProps) {
...
+    <div className="px-8 text-xs text-base-subtle text-center transition-opacity duration-200 data-hidden:opacity-0">
       {entry.mode === "text"
         ? text.switched_to_text_mode
         : text.switched_to_voice_mode}
     </div>
   );
 }
 
-function getMessageComponent(entry: TranscriptEntry, isStreaming?: boolean) {
+function ToolCallMessage({
```
kamilk-11: Nit but the `p` padding from the generated markdown increases the spacing between the agent message that triggered the tool and the tool message:

Normal | Debug
--- | ---
Image | Image

Streamdown adds `*:first:mt-0 *:last:mb-0:` to the markdown container to prevent that. Maybe we should do the same?

PR #465, opened 2026-01-22
[widget] bump to tailwind v4

kalvin807:
Quick effort to use tailwind v4 at widget. allows us to use some modern css without using the opt-out

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -24,7 +24,7 @@ jobs:
       - name: Set up node
         uses: actions/setup-node@v6
         with:
-          node-version: 20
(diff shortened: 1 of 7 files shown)


packages/convai-widget-core/src/styles/index.css
```diff
@@ -109,7 +108,7 @@
     @apply text-base-primary font-medium text-xl;
   }
 
-  .terms h2 {
+  .terms h1 {
```
kalvin807: side fix: from L107, i am quite sure it is a typo

kamilk-11 (approved):

PR #478, opened 2026-01-28
[Widget] fix style in shadow root

kalvin807:
Please help me merge and release if I offlined

## Why
The feature guard at tailwind v4 is bugged in web component. Our doc widget somehow worked most likely due to the doc page itself also using tailwind v4, so the css var get referred from parent.
 
See https://github.com/tailwindlabs/tailwindcss/issues/15005

## What
Add Vite plugin to fix Tailwind CSS issues in Shadow DOM

### Test
I tested by copying the generated css bundle to 11 widget setting page and it worked.

### DIFF
--- a/.changeset/whole-walls-fly.md
+++ b/.changeset/whole-walls-fly.md
@@ -1,0 +1,6 @@
+---
+"@elevenlabs/convai-widget-core": minor
+"@elevenlabs/convai-widget-embed": minor
+---
+
+Fix styling issue in shadow root


packages/convai-widget-core/vite-plugin-tailwind-shadowdom.ts
```diff
@@ -0,0 +1,36 @@
+import type { Plugin } from 'vite';
+
+// Patterns to strip webkit-hyphens media query conditions that cause issues in Shadow DOM
+// See https://github.com/tailwindlabs/tailwindcss/issues/15005
+const WEBKIT_HYPHENS_DOUBLE_PAREN = /\(\(-webkit-hyphens:\s*none\)\)\s*and\s*/g;
+const WEBKIT_HYPHENS_SINGLE_PAREN = /\(-webkit-hyphens:\s*none\)\s*and\s*/g;
+const ROOT_SELECTOR = /:root\b/g;
+const CSS_MODULE_PATTERN = /\.css\?/;
```
AngeloGiacco: How come requires "?"
Do we only use with params

AngeloGiacco (approved):
Lgtm 1q

packages/convai-widget-core/vite-plugin-tailwind-shadowdom.ts
kalvin807: when vite bundle css, it will have `?url` or `?inline` at the end to mark it goes to a file or directly inline in js.

Alletkla:
@kalvin807 
Just saw you mentioned the tailwind issue. I took the same approach like you removing just webkit hyphens, but had problems with safari @work.

If its the case for you as well you can give my 0 dependency plugin a try it removes the whole support block. (Or integrate this in your own plugin)
https://www.npmjs.com/package/vite-plugin-tailwind-shadowdom

PR #484, opened 2026-01-30
fix: Normalisation of the `textOnly` option (in Client package)

kraenhansen:
The top-level `textOnly` option and the  overrides (`overrides.conversation.textOnly`) should do the same thing: Disable mic input and make the agent unreceptive for audio input.

Merging this PR will:
- Normalise across both options, when building up the "full options" object, making their values equal with a preference for the top-level option (and logging a warning) in case they don't match.

### DIFF
--- a/.changeset/bold-sloths-sin.md
+++ b/.changeset/bold-sloths-sin.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/client": patch
+---
+
+Normalize the `textOnly` option (passable both on the top-level and via the overrides object): Providing one will propagate to the other, with the top-level taking precedence, in case of conflict.


packages/client/src/index.test.ts
```diff
@@ -154,7 +154,9 @@ describe("Conversation", () => {
           conversation_config_override: {
             agent: { prompt: { prompt: "CUSTOM_PROMPT" } },
             tts: {},
-            conversation: {},
+            conversation: {
+              text_only: conversationType === "text",
```
kraenhansen: `text_only` is now provided in the event, as the top-level `textOnly` is provided too.

packages/client/src/index.ts
```diff
@@ -69,8 +69,9 @@ export type {
 
 export class Conversation extends BaseConversation {
   public static startSession(options: PartialOptions): Promise<Conversation> {
-    return options.textOnly
-      ? TextConversation.startSession(options)
-      : VoiceConversation.startSession(options);
+    const fullOptions = Conversation.getFullOptions(options);
```
kraenhansen: We're also calling `getFullOptions` inside of `TextConversation.startSession` and `VoiceConversation.startSession`, which could be considered a bit redundant and might warn multiple times 🤔.

Ideally, a future major version would:
- Make `TextConversation` and `VoiceConversation` private to the package and have it's signature updated to take a "full" `Options` object.

PR #487, opened 2026-01-31
add voiceIsolation to WebRTC

AngeloGiacco:
-`voiceIsolation: true` was already enabled for WebSocket connections (in`input.ts`)                                                                 
-WebRTC was missing this because it uses LiveKit's APIs instead of the shared `Input` class                                                         
-This PR adds `voiceIsolation` to WebRTC for both initial mic setup and device switching

### DIFF
--- a/packages/client/src/utils/WebRTCConnection.test.ts
+++ b/packages/client/src/utils/WebRTCConnection.test.ts
@@ -2,22 +2,27 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 
 // Track mock calls using a global object that can be accessed after mocking
 const mockCalls = {
-  setMicrophoneEnabled: [] as boolean[],
+  createLocalAudioTrack: [] as unknown[],
+  publishTrack: [] as unknown[],
 };
 
 vi.mock("livekit-client", () => {
+  const mockAudioTrack = { id: "mock-audio-track" };
+
   const mockLocalParticipant = {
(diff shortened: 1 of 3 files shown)


packages/client/src/utils/WebRTCConnection.test.ts
```diff
@@ -2,22 +2,27 @@ import { describe, it, expect, vi, beforeEach } from "vitest";
 
 // Track mock calls using a global object that can be accessed after mocking
 const mockCalls = {
-  setMicrophoneEnabled: [] as boolean[],
+  createLocalAudioTrack: [] as unknown[],
```
PaulAsjes: Can we use proper types instead of `unknown`?

packages/client/src/utils/WebRTCConnection.ts
```diff
@@ -158,7 +158,18 @@ export class WebRTCConnection extends BaseConnection {
 
       // Enable microphone only if not text-only mode
       if (!config.textOnly) {
-        await room.localParticipant.setMicrophoneEnabled(true);
+        const audioConstraints = {
```
PaulAsjes: Since this is duplicated below, can we extract this? Also are we sure that `setMicrophoneEnabled` doesn't already do this?

PaulAsjes (changes requested):

kraenhansen:
Removed myself as reviewer to avoid reminders, feel free to add me back when / if it's considered ready for review 👍

PR #491, opened 2026-02-03
fix(widget): send feedback when only a conversation id is available

simong:
This fixes the use-case where a widget is rendered using a `signed-url` but no `agent-id` attribute.


kalvin807 (approved):
@simong LGTM. Could you merge the latest changes and generate a patch changelog using Changesets?

PR #522, opened 2026-02-19
Expose conversation connections globally

kai-stra:
This will allow javascript like browser extensions to interact with the conversation, and also allow developers to debug in the console. Currently exposing only what I'm specifically using as although this technically doesn't introduce any malicious capabilities, would prefer to trivialise it less.


kraenhansen (changes requested):
We discussed an alternative approach on Slack

kraenhansen:
Action is on my side to review and move this forward 👍

PR #528, opened 2026-02-24
[Agents] Refactor widget tool status to inline transcript entries

kalvin807:
## Why

The previous implementation #463 tracked tool call states via a separate `Map<eventId, Map<toolCallId, status>>` signal, which was disconnected from the transcript flow. This made it fragile — status had to be derived by cross-referencing the map with transcript entries, and was lost on disconnect since the map was reset.

- Needs: https://github.com/elevenlabs/xi/pull/26328

## What

- Replace Map-based `toolCallStatus` tracking with discrete `agent_tool_request` / `agent_tool_response` transcript entries
- Add `buildDisplayTranscript` utility that computes grouped, filtered display state from raw transcript (message dedup by eventId, tool status derivation, empty message filtering, first message prepend)
  - 📝 technically it forces full re-render per new message, but I think it is okay since chat usually short.
- Propagate `event_id` through `onMessage` and `onAgentChatResponsePart` callbacks
- Make `event_id` required in `TextResponsePart` schema
- Rename "Thinking..." → "Working..." for agent status label
- Tool status now persists after disconnect since it lives in the transcript

## Test I ran
- [x] Unit tests for `buildDisplayTranscript` (grouping, tool status, filtering, edge cases)
- [x] Integration tests for inline tool status rendering (enabled/disabled/post-disconnect)
- [x] Updated client SDK tests for `event_id` propagation

Note: The widget seems to be ignoring some `agent_response` or `tentative_agent_response`. The agent is saying things that never show up in the transcript—usually it skips the preparatory tool-call chatter, e.g "I’m running as foo tool now, plz wait". I’ll fix this in a separate PR.

https://github.com/user-attachments/assets/441af554-adfe-4371-97b1-bb097f9da970

### DIFF
--- a/.changeset/warm-planes-greet.md
+++ b/.changeset/warm-planes-greet.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/convai-widget-core": minor
+---
+
+Add `show-conversation-id` config option to control visibility of conversation ID in disconnection messages. Defaults to `true`. Error messages always show the conversation ID regardless of this setting.
--- a/packages/convai-widget-core/src/contexts/widget-config.tsx
+++ b/packages/convai-widget-core/src/contexts/widget-config.tsx
@@ -103,6 +103,7 @@ export function WidgetConfigProvider({ children }: WidgetConfigProviderProps) {
   const overrideTextOnly = useAttribute("override-text-only");
   const useRtc = useAttribute("use-rtc");
   const showAgentStatus = useAttribute("show-agent-status");
+  const showConversationId = useAttribute("show-conversation-id");
 
   const value = useComputed<WidgetConfig | null>(() => {
     if (!fetchedConfig.value) {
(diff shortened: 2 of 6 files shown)


packages/convai-widget-core/src/widget/Sheet.tsx
```diff
@@ -43,32 +44,19 @@ export function Sheet({ open }: SheetProps) {
...
-        ...transcript.value,
-      ];
-    }
-
-    return config.value.transcript_enabled
-      ? transcript.value
-      : transcript.value.filter(
-          entry => entry.type !== "message" || entry.isText
-        );
+  const filteredTranscript = useComputed<DisplayTranscriptEntry[]>(() => {
```
kalvin807: 📝 I moved all transcript logic to `buildDisplayTranscript`

PR #614, opened 2026-03-20
Create random user id when none is set

lucas-sonnabend:
If no user id is set, create one and store in local storage, so when a user re-visits the page (without refreshing their browser state or incognito session) they are identified as the same user as before

---

> [!NOTE]
> **Medium Risk**
> Adds client-side user identification using FingerprintJS and `localStorage`, which can affect privacy/compliance expectations and introduces a new third-party dependency plus async logic during session start.
> 
> **Overview**
> Ensures conversations always have a stable `userId`: when the widget is started without an explicit `user-id` attribute, it now generates one (preferably via `@fingerprintjs/fingerprintjs`, otherwise `crypto.randomUUID()`), stores it under `elevenlabs_convai_user_id` in `localStorage`, and reuses it on subsequent visits.
> 
> Adds browser tests covering creation, reuse, opt-out when `user-id` is provided, and the FingerprintJS-failure fallback path, and updates deps/lockfile to include FingerprintJS.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 426cd948693acd95f234792ed67f399f91c4cb6c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


kalvin807 (approved):
code lgtm. question on feature
1. do we want to user (the agent owner) a option to opt-in /opt-out persistent tracking
2. maybe server can send event to set the userId instead of relies on fingerprint / session_id

Issue #662, opened 2026-04-09
end_call and max_duration_exceeded silently fail on React Native due to missing browser globals


coryetzkorn:
When an ElevenLabs agent calls the `end_call` system tool, the call does not end on React Native. The `onDisconnect` callback never fires, and the client-side UI remains stuck in a connected state.

The root cause is in `BaseConversation.js`, which uses browser-only globals that don't exist in React Native:

**`handleAgentToolResponse`:**
```js
if (event.agent_tool_response.tool_name === "end_call") {
    this.endSessionWithDetails({
        reason: "agent",
        context: new CloseEvent("end_call", { reason: "Agent ended the call" }),
    });
}
```

**`handleErrorEvent`:**
```js
if (errorType === "max_duration_exceeded") {
    this.endSessionWithDetails({
        reason: "error",
        message: message,
        context: new Event("max_duration_exceeded"),
    });
}
```

`CloseEvent` and `Event` are Web API constructors that don't exist in the React Native JavaScript environment. When these lines execute, they throw a `ReferenceError`:

```
[Error: Uncaught (in promise) ReferenceError: Property 'CloseEvent' doesn't exist]
[Error: Uncaught (in promise) ReferenceError: Property 'Event' doesn't exist]
```

Because the error is thrown before `endSessionWithDetails()` completes, the session is never properly terminated on the client side: `onDisconnect` never fires, timers are never cleared, and the UI stays in the call state indefinitely.

There is also a `document` reference somewhere in the SDK/dependency chain that causes a similar error on connection:
```
[Error: Uncaught (in promise) ReferenceError: Property 'document' doesn't exist]
```

### Impact

- Agent-initiated `end_call` is completely broken on React Native

### Workaround

Polyfill the missing globals on `globalThis` before any ElevenLabs SDK code is imported:

```typescript
const g = globalThis as Record<string, unknown>;

if (typeof g.CloseEvent === "undefined") {
  g.CloseEvent = class CloseEvent {
    type: string;
    reason: string;
    code: number;
    wasClean: boolean;
    constructor(type: string, init?: { reason?: string; code?: number; wasClean?: boolean }) {
      this.type = type;
      this.reason = init?.reason ?? "";
      this.code = init?.code ?? 1000;
      this.wasClean = init?.wasClean ?? true;
    }
  };
}

if (typeof g.Event === "undefined") {
  g.Event = class Event {
    type: string;
    constructor(type: string) {
      this.type = type;
    }
  };
}
```

[...]

PR #671, opened 2026-04-10
Prevent DOM/web API leakage into platform-agnostic code

### LINKED ISSUE #661: DOMException Error on load of app with React Native
ERROR  [ReferenceError: Property 'DOMException' doesn't exist] 

Code: index.tsx
  37 | import { Cog, X } from "lucide-react-native";
  38 | import AsyncStorage from "@react-native-async-storage/async-storage";
> 39 | import { useConversationControls, useConversationStatus } from '@elevenlabs/react-native';
     | ^
  40 |
  41 | import {
  42 |   getSpeechRecognitionLocale,
Call Stack
  <global> (src/app/(app)/(tabs)/(search)/voice_chat/index.tsx:39)

With packages 
npm install @elevenlabs/react-native@1.0.3 @livekit/react-native@2.9.2 @livekit/react-native-webrtc@137.0.2

kraenhansen:
## Summary

Fixes #661 and #662 — React Native crashes caused by browser-only APIs leaking into platform-agnostic code.

- Replace `Event`/`CloseEvent` DOM types in `DisconnectionDetails` with platform-agnostic `DisconnectionEvent`/`DisconnectionCloseEvent` interfaces (structurally compatible, backwards-safe)
- Replace `new CloseEvent()`/`new Event()` browser constructors with plain objects in `BaseConversation.ts`
- Add `readonly type` discriminant (`"webrtc" | "websocket"`) to `BaseConnection`, replacing `instanceof` checks that required value imports of web modules
- Split `VoiceSessionSetup.ts` into core `VoiceSessionStrategy.ts` (zero web imports) and web-only `webSessionSetup.ts`, breaking the eager import chain that caused `livekit-client` to load at module scope in React Native
- Add `@elevenlabs/client/web` entry point for platform-specific code that needs the web session setup
- Restructure RN package imports with a `setup.ts` side-effect module ensuring `registerGlobals()` runs before `livekit-client` loads
- Add ESLint `no-restricted-globals` rules to prevent browser-only globals in core files

## Test plan

- [x] `pnpm turbo build check-types --filter=@elevenlabs/client` passes
- [x] `pnpm turbo build check-types --filter=@elevenlabs/react-native` passes
- [x] `pnpm turbo lint:es --filter=@elevenlabs/client` passes
- [x] `pnpm turbo test --filter=@elevenlabs/client` passes (168/168)
- [ ] Manual: importing `@elevenlabs/react-native` in a React Native Expo project no longer crashes with `DOMException` error
- [ ] Manual: `end_call` and `max_duration_exceeded` events work correctly in React Native


packages/client/eslint.config.mjs
```diff
@@ -1 +1,55 @@
...
+      "no-restricted-syntax": [
+        "error",
+        {
+          selector: "ImportDeclaration[source.value='livekit-client']:not([importKind='type'])",
+          message: "livekit-client must not be imported in core files (causes React Native crashes at module load time). Use a dynamic import() inside an async function instead.",
+        },
+      ],
+    },
+  },
+];
```
kraenhansen: Instead ESLint preventing globals, I'd like to explore a TypeScript based approach with a separate TS project for the DOM related files. Change is coming up.

dragosb0000:
## Real-world confirmation from a production React Native app

We're building a voice coaching app (React Native + Expo SDK 55, `@elevenlabs/react-native@1.0.3`, `@livekit/react-native@2.9.8`) and hit every single issue this PR addresses. Here's the full chain of crashes we encountered, in order:

1. **`context.close is not a function`** — Initially used `signedUrl` which forced WebSocket mode → `MediaDeviceOutput.create()` → `AudioContext` doesn't exist in RN. Switched to `conversationToken` for WebRTC.
2. **`Event` / `CloseEvent` not defined** — `WebRTCConnection.setupRoomEventListeners` creates these in disconnect/connection-state-changed handlers (lines 236, 242).
3. **`window.removeEventListener is not a function`** — `livekit-client`'s `Room#handleDisconnect` uses `window` for visibility/network listeners.
4. **`document.createElement is not a function`** — `Track#attach()` called from `TrackSubscribed` handler (line 271) unconditionally calls `document.createElement('audio')`.
5. **`document.removeEventListener is not a function`** — `Track#removeAppVisibilityListener` during track detach/cleanup.

We're currently shipping with a growing pile of polyfills in our entry point (`DOMException`, `HTMLAudioElement`, `document`, `window`, `Event`, `CloseEvent`) — all stubs just to prevent crashes. The audio actually plays fine through native WebRTC/LiveKit, confirming the DOM code is unnecessary on RN.

We verified the official example at `examples/react-native-expo` has zero polyfills, which initially confused us — but it works because the example runs against `workspace:*` (unreleased code). Published `@elevenlabs/client@1.1.2` still has all the DOM calls.

**This PR is exactly the right fix.** The approach of splitting web-only code out of the shared path and adding ESLint guards is solid. Would love to see this merged and published — happy to test a canary/prerelease if that helps move it along.

**Our stack:** React Native 0.83.2, Expo SDK 55, `@elevenlabs/react-native@1.0.3`, `@elevenlabs/client@1.1.2`, `@livekit/react-native@2.9.8`, `livekit-client@2.18.1`

asim48-ctrl:
I took a quick current-`main` release-readiness pass on this because #766 is now confirmed and the remaining risk looks like merge/release friction rather than diagnosis.

What I see as of today:

- The PR is still `DIRTY` against `main`.
- `git merge-tree origin/main pr-671` shows a real content conflict in `packages/client/package.json`; `packages/client/src/BaseConversation.ts` and `packages/client/src/VoiceConversation.ts` auto-merge, but they are in the affected set and worth rechecking after the rebase.
- The branch is stale enough that `packages/client/package.json` would move the client version from current `1.7.0` back to `1.1.2` in a raw `main..pr-671` diff.
- The diff has a lot of broad churn outside the core RN/DOM fix path: changelog removals, example deletions, widget/react/test/scribe changes, lock/workspace changes, etc. That makes the actual useful patch harder to review than it needs to be.

The core fix direction still looks right to me:

- split `webSessionSetup` out of the platform-agnostic client path;
- stop constructing DOM `Event`/`CloseEvent` in shared conversation code;
- use a `connection.type` discriminant instead of `instanceof` checks that pull web values into RN;
- make the RN package register globals before LiveKit/web-adjacent code can load.

If I were trying to land this with the least reviewer friction, I would rebase it as a smaller patch set around only these surfaces:

- `packages/client/src/platform/VoiceSessionStrategy.ts`
- `packages/client/src/platform/webSessionSetup.ts`
- `packages/client/src/BaseConversation.ts`
- `packages/client/src/VoiceConversation.ts`
- `packages/client/src/utils/BaseConnection.ts`
- `packages/client/src/utils/WebRTCConnection.ts`
- `packages/client/src/utils/WebSocketConnection.ts`
- `packages/client/src/index.ts`
- `packages/client/src/internal.ts`
- `packages/react-native/src/setup.ts`
- `packages/react-native/src/index.react-native.ts`
- `packages/react-native/src/nativeVolume.ts`
- the relevant type definitions + package exports + changeset

Then validate against the actual external-user path, not just the workspace example: fresh Expo SDK 54 / RN 0.81 app, published `@elevenlabs/react-native@1.2.0` baseline reproduces, rebased branch or packed canary no longer hits the import/start-session crash path.

The important distinction from #766 is that this does not need more root-cause analysis. It needs a clean rebase/reduction so the fix can ship without accidentally dragging older package metadata and unrelated workspace churn back across `main`.

asim48-ctrl:
I finished the external-user iOS repro pass against the published packages.

Fresh Expo/iOS simulator app, built on a macOS runner from npm packages only:

- `@elevenlabs/react-native@1.2.0`
- `@livekit/react-native@2.9.8`
- Hermes / iOS simulator

The app builds and launches, then the published RN path hits the browser media surface at runtime:

```text
[EL_REPRO] start direct media path
[EL_REPRO] direct media path failed navigator.mediaDevices.getSupportedConstraints is not a function (it is undefined)
```

The stack is in the bundled ElevenLabs client path on iOS/Hermes. The same run also gets the public `Conversation.startSession({ connectionType: "websocket" })` path as far as `status: "connecting"`, so the app/runtime is alive rather than failing at install or launch.

Runner artifact:

- https://github.com/asim48-ctrl/elevenlabs-rn-repro-runner/actions/runs/25894684051

This lines up with the earlier production report here: the published RN package still lets browser media/DOM assumptions leak into the external-user React Native path. My current read is unchanged: #671 has the right direction, but the useful next step is a reduced/rebased patch plus validation against a fresh app using packed/published-package shape, not only the workspace example.

kraenhansen:
Closing this as I'm working on a different approach over in https://github.com/elevenlabs/packages/pull/784

PR #672, opened 2026-04-11
fix(react): fire onConnect after conversationRef is set

kalvin807:
# Why

`onConnect` and `connected` currently describe two different readiness points.

On `main`, the client fires `onConnect` before the session is marked `connected`, and React only assigns `conversationRef` after `Conversation.startSession()` resolves. That means React consumers can hit `No active conversation` if they call controls from `onConnect`, while direct client readiness checks like `isOpen()` still lag behind message-sending behavior.

This change aligns the lifecycle so `connected` means ready, and makes the React provider participate in the same lifecycle instead of replaying `onConnect` later.

```mermaid
sequenceDiagram
    participant ReactUser as "ReactUser onConnect"
    participant Provider as "ConversationProvider"
    participant Client as "Conversation.startSession"
    participant Session as "Conversation"

    rect rgb(0,0,0)
        Note over ReactUser,Session: main
        ReactUser->>Provider: startSession(options)
        Provider->>Client: startSession(sessionOptionsWithUserOnConnect)
        Client->>Session: construct conversation
        Session->>ReactUser: onConnect({ conversationId })
        Note over Provider: conversationRef still null
        Note over Session: status still connecting
        Session-->>Session: set status connected
        Client-->>Provider: resolve(conversation)
        Provider->>Provider: conversationRef = conversation
    end

    rect rgb(0,0,0)
        Note over ReactUser,Session: final
        ReactUser->>Provider: startSession(options)
        Provider->>Client: startSession(wrappedOnConnect + onConversationCreated)
        Client->>Session: construct conversation
        Session->>Provider: onConversationCreated(conversation)
        Provider->>Provider: conversationRef = conversation
        Session-->>Session: set status connected
        Session->>ReactUser: onConnect({ conversationId })
    end
```

# What

- Refactor client session startup so concrete `startSession()` flows own the `connected` and `onConnect` ordering instead of `BaseConversation` constructors.
- Mark the session `connected` before user `onConnect`, so `isOpen()` and callback-time readiness agree.
- Let the React provider synchronize `conversationRef` through the same lifecycle rather than stripping and replaying `onConnect` after promise resolution.
- Add regressions covering:
  - calling controls from React `onConnect`
  - thrown `onConnect` flowing through the normal error path
  - the new client ordering of `connected` before `onConnect`
  - startup cleanup preserving the original error when teardown also throws

---

> [!NOTE]
> **Medium Risk**
> Changes conversation session startup ordering and React provider lifecycle handling, which can affect when consumer code runs and how errors/cleanup propagate. Risk is mitigated by added regression tests covering ordering, error paths, and stale-session handling.
> 
> **Overview**
> **Fixes conversation readiness semantics** by moving `onConnect` out of `BaseConversation` construction and ensuring `TextConversation`/`VoiceConversation` call `markConnected()` *before* invoking user `onConnect`, with improved cleanup when startup fails.
> 
> **Adds a new lifecycle callback** `onConversationCreated(conversation)` to the client options/types and exports it (including internal exports), then updates the React `ConversationProvider` to set `conversationRef`/state via `onConversationCreated`, wrap `onConnect`, and ignore stale async resolutions via a monotonic `startSessionId`.
> 
> **Expands test coverage** in both packages to verify `connected` precedes `onConnect`, React consumers can safely call controls in `onConnect`, thrown `onConnect` transitions to error/clears conversation, stale start/restart scenarios are ignored, and original startup errors are preserved even if cleanup throws.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 891e6316f1e2cb911c75d04bce35d0311ca614ca. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/react/src/conversation/ConversationProvider.test.tsx
+++ b/packages/react/src/conversation/ConversationProvider.test.tsx
@@ -706,6 +706,64 @@ describe("ConversationProvider", () => {
       mockConversation
     );
   });
+
+  it("ignores stale startSession resolution after restarting inside onConnect", async () => {
+    const firstConversation = createMockConversation("first-id");
+    const secondConversation = createMockConversation("second-id");
+    const { promise: firstPromise, resolve: resolveFirst } =
+      Promise.withResolvers<Conversation>();
+    const { promise: secondPromise, resolve: resolveSecond } =
+      Promise.withResolvers<Conversation>();
+    let firstOptions: MockStartSessionOptions | null = null;
+    let secondOptions: MockStartSessionOptions | null = null;
+
+    vi.mocked(Conversation.startSession).mockImplementation(options => {
+      if (!firstOptions) {
+        firstOptions = options as MockStartSessionOptions;
+        return firstPromise;
+      }
+      secondOptions = options as MockStartSessionOptions;
+      return secondPromise;
+    });
+
+    const { result } = renderHook(() => useTestContext(), {
+      wrapper: createWrapper(),
+    });
+
+    act(() => {
+      result.current.startSession({
+        onConnect: () => {
+          result.current.endSession();
+          result.current.startSession();
+        },
+      });
+    });
+
+    act(() => {
+      driveConnectedSessionLifecycle(firstOptions!, firstConversation);
+    });
+
+    expect(Conversation.startSession).toHaveBeenCalledTimes(2);
+    expect(result.current.conversation).toBeNull();
+
+    act(() => {
+      driveConnectedSessionLifecycle(secondOptions!, secondConversation);
+    });
+
+    expect(result.current.conversation).toBe(secondConversation);
+
+    await act(async () => {
+      resolveFirst(firstConversation);
+    });
+
+    expect(result.current.conversation).toBe(secondConversation);
+
+    await act(async () => {
+      resolveSecond(secondConversation);
+    });
+
+    expect(result.current.conversation).toBe(secondConversation);
+  });
 });
 
 describe("CALLBACK_KEYS", () => {
--- a/packages/react/src/conversation/ConversationProvider.test.tsx
+++ b/packages/react/src/conversation/ConversationProvider.test.tsx
@@ -764,6 +764,59 @@ describe("ConversationProvider", () => {
 
     expect(result.current.conversation).toBe(secondConversation);
   });
+
+  it("ignores stale startSession rejection after restarting inside onConnect", async () => {
+    const onError = vi.fn();
+    const firstConversation = createMockConversation("first-id");
+    const secondConversation = createMockConversation("second-id");
+    const firstError = new Error("first failed");
+    const { promise: firstPromise, reject: rejectFirst } =
+      Promise.withResolvers<Conversation>();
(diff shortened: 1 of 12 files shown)


packages/react/src/conversation/ConversationProvider.test.tsx
```diff
@@ -575,14 +565,58 @@ describe("ConversationProvider", () => {
       result.current.startSession();
     });
 
-    // Verify Conversation.startSession was called with the provided callback
     const startSessionCall = vi.mocked(Conversation.startSession).mock
       .calls[0][0];
-    expect(typeof startSessionCall.onConnect).toBe("function");
+    // onConnect is stripped from options and fired by the provider after
+    // startSession resolves, so it should NOT be on the options object
+    expect(startSessionCall.onConnect).toBeUndefined();
```
kraenhansen: My initial reaction is that this is not correct 🤔 
We're preventing the client control of the callback, which might change semantics and also makes it harder to introduce additive changes to the arguments in the future.

Not strongly against this, but I'd like to share an alternative, of calling `startSession` with an `onConnect` created by the provider, setting the ref and calling the callback provided by the user. Makes sense? Then we preserve the semantics controlled by the client, we can pass the arguments directly on instead of having to recreate them and the ref will be set before the users callback is called.

I think this will also have the added benefit of propagating any errors thrown from `onConnect` to the regular error path, setting the provider into an error state - which we should also test for.

kraenhansen (approved):
Not strongly against this, but I've shared an alternative in my inline comment. I can update with the alternative suggestion in a comment if my alternative proposal is unclear.

kalvin807:
Humm not something I imagined. Let me manually fix it later

packages/react/src/conversation/ConversationProvider.tsx
```diff
@@ -145,20 +153,62 @@ export function ConversationProvider({
       clientToolsRef.current = clientTools;
       sessionOptions.clientTools = clientTools;
 
-      lockRef.current = Conversation.startSession(sessionOptions);
+      const userOnConversationCreated = sessionOptions.onConversationCreated;
+      const isStaleStartSession = () =>
+        startSessionId !== startSessionIdRef.current;
```
kraenhansen, line 158: As we can't rely on the ordering of these async operations, I fear this could potentially deadlock 🤔 I'm not absolutely certain - perhaps explicit tests around that (resolving the first conversation after the second) would suffice.

I'm also left wondering why we can't simply rely on object identity here (`conversationRef.current === conversation` or similar) instead of using numeric values to represent id here.

kraenhansen (approved):
I agree, having a separate `onConversationCreated` hook is cleaner. I've got a suggestion and some questions but those could be handled in follow-ups if needed 👍

packages/react/src/conversation/ConversationProvider.tsx
kalvin807: conversationRef.current === conversation only helps when a conversation instance exists. Stale .catch(...) callbacks do not have one, so we still need attempt-level tracking.

PR #675, opened 2026-04-14
feat(scribe): add native mute/unmute support

### LINKED ISSUE #674: Support a native mute/unmute audio functionality for Scribe v2 realtime STT integration
### Problem

I'm building a voice agent utilizing Scribe v2 realtime as the STT component. In one of our use cases, users often pause the conversation to perform tasks elsewhere on the screen. During this pause, if a user inadvertently speaks, the STT will pick up and process that audio, which is undesirable.

Currently, we handle this by managing a mute/unmute flag and ignoring transcripts (both partial and committed) when muted. However, this is more of a workaround. If we unmute in the middle of a user's utterance, the STT still returns the entire audio—resulting in unwanted transcriptions even though the user was muted partway through speaking.

A typical solution would involve directly controlling the microphone node and disconnecting it on mute, but this requires us to handle microphone management, which is better abstracted within the Scribe SDK.

### Feature Request

It would be ideal if the Scribe SDK or STT integration provided a native mute/unmute option—so that when muted, it does not process or send audio, and when unmuted, it resumes seamlessly. This would greatly simplify integration and avoid transcription of unintended speech.

**Is this feature feasible to add, or is there guidance for proper handling of mute/unmute scenarios with the Scribe stack?**

hardiksondagar:
## Summary

Fixes #674.

Adds native mute/unmute functionality to the Scribe realtime STT integration so callers no longer need to manage a client-side ignore-flag workaround.

- **`RealtimeConnection.mute()`** — disables the underlying `MediaStreamTrack` (microphone mode) so the browser stops delivering real audio to the worklet, and sets an internal `_muted` flag so `send()` becomes a no-op (manual mode)
- **`RealtimeConnection.unmute()`** — re-enables the track and clears the flag, resuming audio transmission
- **`RealtimeConnection.isMuted`** — boolean getter reflecting current mute state
- **`useScribe` hook** — exposes `isMuted` state, `mute()`, and `unmute()` callbacks

### How it works

```
mute()
  ├── track.enabled = false   → browser captures silence (mic mode)
  └── _muted = true           → send() becomes a silent no-op (both modes)

unmute()
  ├── track.enabled = true    → real mic audio resumes
  └── _muted = false          → send() forwards audio again
```

The worklet `onmessage` handler also guards on `isMuted` as a second layer, ensuring no audio chunks slip through even if the browser takes a moment to honor `track.enabled = false`.

## Test plan

- [x] All 56 existing scribe tests pass (28 Chromium + 28 Firefox)
- [x] 4 new `Mute / Unmute` tests added and passing in both browsers:
  - starts unmuted by default
  - `mute()` / `unmute()` toggle `isMuted`
  - `send()` is a no-op while muted
  - `send()` resumes after `unmute()`
- [x] TypeScript check passes for `@elevenlabs/client` and `@elevenlabs/react`
- [x] ESLint passes for all changed files

Made with [Cursor](https://cursor.com)

---

> [!NOTE]
> **Medium Risk**
> Moderate risk because it changes realtime audio capture behavior (toggling `MediaStreamTrack.enabled`) and adds new public API surface in both the client and React hook, which could affect streaming edge cases and connection lifecycle.
> 
> **Overview**
> Adds native mute/unmute support to Scribe realtime STT: `RealtimeConnection` now tracks `isMuted` and provides `mute()`/`unmute()` that disable/enable the underlying microphone `MediaStreamTrack` when streaming from mic.
> 
> Updates microphone setup to persist the captured track on the connection (and apply any pre-existing muted state), documents that `send()` still transmits while muted, and extends the React `useScribe` hook to expose `isMuted` plus `mute()`/`unmute()` (resetting mute state on disconnect/close). Includes new unit tests covering the mute state toggling and the fact that `send()` continues to forward audio while muted.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1596eaab1fe7fdf24d03e9f5922128dd113e8216. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/scribe/connection.ts
+++ b/packages/client/src/scribe/connection.ts
@@ -197,10 +197,57 @@ export class RealtimeConnection {
   private websocket: WebSocket | null = null;
   private eventEmitter: EventEmitter = new EventEmitter();
   private currentSampleRate: number = 16000;
+  private _muted: boolean = false;
   public _audioCleanup?: () => void;
+  /** @internal Set by ScribeRealtime in microphone mode to enable track-level muting. */
+  public _mediaStreamTracks?: MediaStreamTrack[];
 
   constructor(sampleRate: number) {
     this.currentSampleRate = sampleRate;
+  }
+
+  /**
+   * Whether audio sending is currently muted.
+   *
+   * In microphone mode, the underlying `MediaStreamTrack` is disabled so the
+   * browser captures silence instead of real microphone input.
+   * In manual mode, calls to `send()` are silently discarded while muted.
+   */
+  public get isMuted(): boolean {
+    return this._muted;
+  }
+
+  /**
+   * Mutes audio capture and transmission.
+   *
+   * In microphone mode, disables the underlying `MediaStreamTrack` so the browser
+   * stops delivering real audio to the worklet. Combined with the worklet-level
+   * guard in `scribe.ts`, no audio chunks reach the server while muted.
+   *
+   * In manual mode, subsequent `send()` calls are silently discarded until `unmute()`.
+   */
+  public mute(): void {
+    this._muted = true;
+    if (this._mediaStreamTracks) {
+      for (const track of this._mediaStreamTracks) {
+        track.enabled = false;
+      }
+    }
+  }
+
+  /**
+   * Unmutes audio capture and transmission.
+   *
+   * Re-enables the `MediaStreamTrack` (microphone mode) or allows `send()` calls
+   * to pass through again (manual mode).
+   */
+  public unmute(): void {
+    this._muted = false;
+    if (this._mediaStreamTracks) {
+      for (const track of this._mediaStreamTracks) {
+        track.enabled = true;
+      }
+    }
   }
 
   /**
@@ -419,6 +466,10 @@ export class RealtimeConnection {
   }): void {
     if (!this.websocket || this.websocket.readyState !== WebSocket.OPEN) {
       throw new Error("WebSocket is not connected");
+    }
+
+    if (this._muted) {
+      return;
     }
 
     const message: InputAudioChunk = {
--- a/packages/client/src/scribe/connection.ts
+++ b/packages/client/src/scribe/connection.ts
@@ -207,24 +207,26 @@ export class RealtimeConnection {
   }
 
   /**
-   * Whether audio sending is currently muted.
+   * Whether audio capture is currently muted.
    *
    * In microphone mode, the underlying `MediaStreamTrack` is disabled so the
-   * browser captures silence instead of real microphone input.
-   * In manual mode, calls to `send()` are silently discarded while muted.
+   * browser captures silence instead of real microphone input. Silence continues
+   * to be sent to the server to keep the connection alive.
    */
   public get isMuted(): boolean {
     return this._muted;
   }
 
   /**
-   * Mutes audio capture and transmission.
+   * Mutes audio capture.
    *
    * In microphone mode, disables the underlying `MediaStreamTrack` so the browser
-   * stops delivering real audio to the worklet. Combined with the worklet-level
-   * guard in `scribe.ts`, no audio chunks reach the server while muted.
+   * replaces real microphone input with silence. The silence continues to flow to
+   * the server, keeping the connection alive without producing transcriptions.
    *
-   * In manual mode, subsequent `send()` calls are silently discarded until `unmute()`.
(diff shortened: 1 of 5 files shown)


packages/client/src/scribe/scribe.ts
```diff
@@ -287,8 +287,20 @@ export class ScribeRealtime {
         });
       }
 
+      // Store tracks so mute()/unmute() can toggle track.enabled.
+      // If mute() was called before mic setup finished, honour that state now.
+      connection._mediaStreamTracks = stream.getAudioTracks();
```
kraenhansen: My instinct tells me that we can expect just a single audio track here. That seems to be the assumption when initialising `trackSettings` above as well. Instead of storing a list here, I think we could simplify and store just a single track and instead of storing a local `_muted` property, I think we should derive that from the track's `enabled` state directly.

packages/client/src/scribe/connection.ts
```diff
@@ -421,6 +468,10 @@ export class RealtimeConnection {
       throw new Error("WebSocket is not connected");
     }
 
+    if (this._muted) {
+      return;
+    }
```
kraenhansen: This seems redundant if the track is already being disabled. Is it intentional that you don't want to send silence here? I suspect this could result in a timeout in the communication with the server and we might want to send zeros instead 🤔

packages/client/src/scribe/scribe.ts
hardiksondagar: Updated to _mediaStreamTrack (singular) in e2126bb. Kept the _muted boolean though — in manual mode there's no MediaStreamTrack to derive from.

packages/client/src/scribe/connection.ts
hardiksondagar: Agreed — removed both guards in 2cc6dbe. track.enabled = false already produces silence, so we let it flow through to keep the connection alive.

hardiksondagar: This could create ambiguity for manual-mode users, so I've added explicit JSDoc on isMuted, mute(), unmute(), and send() clarifying that mute controls SDK microphone capture only, and send() is intentionally never blocked. See eec9adb.

hardiksondagar:
@kraenhansen anything else to add here?

packages/react/src/scribe.ts
```diff
@@ -191,6 +196,7 @@ export function useScribe(options: ScribeHookOptions = {}): UseScribeReturn {
   const connectionRef = useRef<RealtimeConnection | null>(null);
 
   const [status, setStatus] = useState<ScribeStatus>("disconnected");
+  const [isMuted, setIsMuted] = useState<boolean>(false);
```
kraenhansen, line 199: Why keep a separate local state instead of simply propagating the internal state of the connection? Is this mainly to trigger a re-render?

packages/client/src/scribe/connection.ts
```diff
@@ -197,12 +197,59 @@ export class RealtimeConnection {
...
 
+  /**
+   * Whether audio capture is currently muted.
+   *
+   * In microphone mode, the underlying `MediaStreamTrack` is disabled so the
+   * browser captures silence instead of real microphone input. Silence continues
+   * to be sent to the server to keep the connection alive.
+   *
+   * In manual mode, this flag is informational only and does not automatically
+   * block `send()`.
```
kraenhansen, line 217: I don't think "microphone" vs "manual" mode is clear here, are you referring to the commit strategies?

hardiksondagar: What i meant, audio coming from non microphone sources. Like file or sending direct base64.

kraenhansen: Okay, I don't think that's clear and it also conflicts with the term "manual" we're using for commit strategy.

packages/client/src/scribe/scribe.ts
kraenhansen: If setting this when there's no media stream has no effect, I think we should simply throw if either of these are called. Perhaps have `isMuted` return false if no media stream track is available, since throwing from a getter is annoying.

packages/client/src/scribe/connection.ts
hardiksondagar: Okay, let me update the comment. I agree, it may conflict with the commit strategy.

kraenhansen (approved):

PR #681, opened 2026-04-17
Add `uploadConversationFile`

bigbernnn:
Adds `uploadConversationFile` to `BaseConversation`, enabling client-side file uploads during active conversations via POST `/v1/convai/conversations/{id}/files`. The method returns a `{ fileId }` that can be passed to `sendMultimodalMessage`. 

Also wired through `useConversationControls()` in the React hooks, which surfaces it automatically in both `@elevenlabs/react` and `@elevenlabs/react-native`. The example app demonstrates the flow.

---

> [!NOTE]
> **Medium Risk**
> Adds a new network upload path (`fetch` + `FormData`) tied to active conversation IDs, which can fail due to auth/origin/CORS and introduces new error-handling paths. Surface area is moderate since it’s additive and covered by basic unit tests and an example integration.
> 
> **Overview**
> Enables **client-side file uploads within an active conversation** by adding `BaseConversation.uploadConversationFile(file: Blob)`, which POSTs to `/v1/convai/conversations/{conversationId}/files` and returns `{ fileId }` for use with `sendMultimodalMessage`.
> 
> Wires this API through `useConversationControls()` (React hooks) and exports the new `UploadConversationFileResult` type, plus adds unit tests for origin scheme normalization and filename generation. Updates the React Native Expo example to pick an image via `expo-image-picker`, upload it, and send a multimodal message with an uploading UI/preview.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 7e6d06a5ab750b34969087815bfc5fafa67f467b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/react-native-expo/App.tsx
+++ b/examples/react-native-expo/App.tsx
@@ -7,7 +7,10 @@ import {
   TouchableOpacity,
   Keyboard,
   TextInput,
+  Image,
+  ActivityIndicator,
 } from "react-native";
+import * as ImagePicker from "expo-image-picker";
 import { VolumeBar } from "./VolumeBar";
 import { FrequencyBands } from "./FrequencyBands";
 import {
@@ -37,9 +40,13 @@ const ConversationScreen = () => {
     sendUserMessage,
     sendContextualUpdate,
     sendUserActivity,
+    sendMultimodalMessage,
+    uploadConversationFile,
     getId,
   } = useConversationControls();
   const isStarting = status === "connecting";
+  const [isUploading, setIsUploading] = useState(false);
+  const [selectedImage, setSelectedImage] = useState<string | null>(null);
 
   const handleSubmitText = () => {
     if (textInput.trim()) {
@@ -65,6 +72,30 @@ const ConversationScreen = () => {
       endSession();
     } catch (error) {
       console.error("Failed to end conversation:", error);
+    }
+  };
+
+  const handlePickAndUploadImage = async () => {
+    const result = await ImagePicker.launchImageLibraryAsync({
+      mediaTypes: ["images"],
+      quality: 0.8,
+    });
+    if (result.canceled) return;
+
+    const asset = result.assets[0];
+    setSelectedImage(asset.uri);
+    setIsUploading(true);
+
+    try {
+      const blob = await fetch(asset.uri).then(r => r.blob());
+      const { fileId } = await uploadConversationFile(blob);
+      sendMultimodalMessage({ fileId, text: textInput.trim() || undefined });
+      setTextInput("");
+    } catch (error) {
+      console.error("Failed to upload image:", error);
+    } finally {
+      setIsUploading(false);
+      setSelectedImage(null);
     }
   };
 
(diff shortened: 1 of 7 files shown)


packages/client/src/BaseConversation.ts
```diff
@@ -580,4 +586,34 @@ export abstract class BaseConversation {
...
+  ): Promise<UploadConversationFileResult> {
+    const origin = (this.options.origin ?? HTTPS_API_ORIGIN).replace(
+      /^wss:\/\//,
+      "https://"
+    );
+
+    const filename =
+      "name" in file && typeof file.name === "string"
+        ? file.name
+        : `upload.${(file.type || "image/png").split("/").pop()}`;
```
PennyroyalTea: what if file type is something like `image/svg+xml`

packages/client/src/BaseConversation.ts
```diff
@@ -580,4 +586,34 @@ export abstract class BaseConversation {
         : undefined,
     });
   }
+
+  public async uploadConversationFile(
+    file: Blob
+  ): Promise<UploadConversationFileResult> {
+    const origin = (this.options.origin ?? HTTPS_API_ORIGIN).replace(
+      /^wss:\/\//,
```
PennyroyalTea: don't think we have `ws://` but this would fail it, maybe worth being on defensive side

PennyroyalTea:
Left two comments, but looks good otherwise. Given no review from subject matter experts, could you double down on testing before shipping? like test e2e and perhaps some unit / integration tests?

examples/react-native-expo/App.tsx
```diff
@@ -299,6 +330,32 @@ const ConversationScreen = () => {
               <Text style={styles.buttonText}>Send Context</Text>
             </TouchableOpacity>
           </View>
+
+          <TouchableOpacity
```
kraenhansen, line 334: Mind moving this into a separate component with its own file and imports?

packages/client/src/BaseConversation.ts
```diff
@@ -580,4 +586,34 @@ export abstract class BaseConversation {
...
+    const body = new FormData();
+    body.append("file", file, filename);
+
+    const response = await fetch(
+      `${origin}/v1/convai/conversations/${this.connection.conversationId}/files`,
+      { method: "POST", body }
+    );
+
+    if (!response.ok) {
+      const text = await response.text().catch(() => "");
```
kraenhansen, line 611: Won't this endpoint return a JSON error object on failures?

packages/client/src/BaseConversation.ts
```diff
@@ -580,4 +586,34 @@ export abstract class BaseConversation {
         : undefined,
     });
   }
+
+  public async uploadConversationFile(
```
kraenhansen, line 590: Since this is off the conversation class, the "conversation" part of this method name seems redundant to me.

```suggestion
  public async uploadFile(
```

The same goes for the names used on the React package API.

packages/client/src/BaseConversation.ts
```diff
@@ -580,4 +586,34 @@ export abstract class BaseConversation {
...
+      `${origin}/v1/convai/conversations/${this.connection.conversationId}/files`,
+      { method: "POST", body }
+    );
+
+    if (!response.ok) {
+      const text = await response.text().catch(() => "");
+      throw new Error(`Upload failed: ${response.status} ${text}`);
+    }
+
+    const { file_id } = await response.json();
```
kraenhansen, line 615: We should assert the value of this `file_id` and its type before returning it.

packages/client/src/BaseConversation.ts
```diff
@@ -580,4 +586,34 @@ export abstract class BaseConversation {
         : undefined,
     });
   }
+
+  public async uploadConversationFile(
+    file: Blob
+  ): Promise<UploadConversationFileResult> {
+    const origin = (this.options.origin ?? HTTPS_API_ORIGIN).replace(
```
kraenhansen: Instead of moving this logic into every use of the options, it'd rather that we resolve these defaults once and rely on the (non-partial / fully qualified) options here.

I started that pattern with the `textOnly` field already.

PennyroyalTea (approved):

kraenhansen:
@bigbernnn I made some comments yesterday but it seems they remained in a "pending" state. I have some changes that I'd like you to consider before releasing this in a regular minor version (a pre-release as-is might be fine).

bigbernnn:
> @bigbernnn I made some comments yesterday but it seems they remained in a "pending" state. I have some changes that I'd like you to consider before releasing this in a regular minor version (a pre-release as-is might be fine).

Following on this here: https://github.com/elevenlabs/packages/pull/683

PR #683, opened 2026-04-21
Extract `ImageUpload` component, rename `uploadFile`, and improve upload error handling

bigbernnn:
> [!NOTE]
> **Medium Risk**
> Introduces a breaking API rename (`uploadConversationFile` -> `uploadFile`) and changes error handling/response validation around network calls, which may affect downstream integrations if not updated.
> 
> **Overview**
> Adds first-class support for uploading files to conversations by renaming the API from `uploadConversationFile` to `uploadFile` across `@elevenlabs/client` and `@elevenlabs/react`, including updated exported types.
> 
> Improves robustness and debuggability by introducing `extractApiErrorMessage()` for consistent API error extraction (used in file upload + WebRTC token fetch) and by validating `file_id` in upload responses.
> 
> Updates the React Native Expo example to use the new `uploadFile` API and extracts the image upload UI/logic into a dedicated `ImageUpload` component.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 810d17ce1dd0b660c93325893a5970fbf3c86337. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/BaseConversation.ts
+++ b/packages/client/src/BaseConversation.ts
@@ -590,10 +590,9 @@ export abstract class BaseConversation {
   public async uploadConversationFile(
     file: Blob
   ): Promise<UploadConversationFileResult> {
-    const origin = (this.options.origin ?? HTTPS_API_ORIGIN).replace(
-      /^wss?:\/\//,
-      "https://"
-    );
+    const origin = (this.options.origin ?? HTTPS_API_ORIGIN)
+      .replace(/^wss:\/\//, "https://")
+      .replace(/^ws:\/\//, "http://");
 
     const filename =
       "name" in file && typeof file.name === "string"
--- a/packages/client/src/BaseConversation.ts
+++ b/packages/client/src/BaseConversation.ts
@@ -126,7 +126,6 @@ export abstract class BaseConversation {
       onStatusChange: () => {},
       onCanSendFeedbackChange: () => {},
       onInterruption: () => {},
-      origin: HTTPS_API_ORIGIN,
       ...partialOptions,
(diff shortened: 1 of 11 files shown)


packages/client/src/BaseConversation.ts
```diff
@@ -609,10 +607,23 @@ export abstract class BaseConversation {
...
-      throw new Error(`Upload failed: ${response.status} ${text}`);
+      let message = text;
+      try {
+        const body = JSON.parse(text);
+        const detail = body?.detail?.message ?? body?.detail;
+        if (typeof detail === "string") {
+          message = detail;
+        }
+      } catch (_) {
+        // Response isn't JSON, use raw text
```
kraenhansen: Not a fan of catch blocks swallowing errors: I've been bit by this too many times already. At the very least a console.warn with the error argument.

I also think we can expect more from our API: Use `json()` on the response and expect the message to be available in the object in exactly one path or not at all.

Finally, the extraction of errors from a response also seems like a shared utility that we'd like to break out for reuse (if we don't already have that in the client package elsewhere).

kraenhansen (approved):
I'm okay with this merging and releasing as is. I've got a few suggestions on how to improve it further.

packages/client/src/BaseConversation.ts
bigbernnn: that's fair. Adding a warning log and extract logic to an error

kraenhansen (approved):

PR #692, opened 2026-04-26
[Fix] - Cancel fade out if new agent audio arrives before timeout

Hashir-Majeed:
> [!NOTE]
> **Medium Risk**
> Touches real-time audio playback interruption/fade logic and changes when output buffers are flushed, which can impact perceived audio glitches or stuck states but does not affect auth or persisted data.
> 
> **Overview**
> Fixes voice playback so an `interrupt()` fade-out is cancelled when new agent audio arrives: `MediaDeviceOutput.playAudio()` now clears scheduled gain ramps, restores the configured volume, clears any pending interrupt timeout, and resets the worklet’s interrupted state before buffering new audio.
> 
> Simplifies interruption handling in `VoiceConversation.handleInterruption()` to always switch back to `listening` and interrupt output (removing the prior event-id-based skip), and removes the corresponding unit test that asserted the skip behavior.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 97887d8354decc3b04867a39f3269c9f95d9f999. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/utils/output.ts
+++ b/packages/client/src/utils/output.ts
@@ -169,7 +169,13 @@ export class MediaDeviceOutput
   }
 
   public playAudio(chunk: ArrayBuffer): void {
-    if (this.interrupted) return;
+    this.gain.gain.cancelScheduledValues(this.context.currentTime);
+    this.gain.gain.value = this.volume;
+    if (this.interruptTimeout) {
+      clearTimeout(this.interruptTimeout);
+      this.interruptTimeout = null;
+    }
+    this.worklet.port.postMessage({ type: "clearInterrupted" });
     this.worklet.port.postMessage({ type: "buffer", buffer: chunk });
   }
 
--- a/packages/client/src/VoiceConversation.ts
+++ b/packages/client/src/VoiceConversation.ts
@@ -169,14 +169,8 @@ export class VoiceConversation extends BaseConversation {
 
   protected override handleInterruption(event: InterruptionEvent) {
     super.handleInterruption(event);
-
-    const interruptId = event.interruption_event?.event_id;
-    const skipOutputFlush =
-      interruptId !== undefined && interruptId < this.currentEventId;
-    if (!skipOutputFlush) {
-      this.updateMode("listening");
-      this.output.interrupt();
-    }
+    this.updateMode("listening");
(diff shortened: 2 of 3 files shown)


packages/client/src/utils/output.ts
```diff
@@ -169,7 +169,13 @@ export class MediaDeviceOutput
   }
 
   public playAudio(chunk: ArrayBuffer): void {
-    if (this.interrupted) return;
+    this.gain.gain.cancelScheduledValues(this.context.currentTime);
```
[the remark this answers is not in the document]
Hashir-Majeed: indeed it is, maybe we should remove it altogether - I don't see any other place we use this other than this early guard we are removing now

packages/client/src/utils/output.ts
```diff
@@ -169,7 +169,13 @@ export class MediaDeviceOutput
...
+    this.gain.gain.value = this.volume;
+    if (this.interruptTimeout) {
+      clearTimeout(this.interruptTimeout);
+      this.interruptTimeout = null;
+    }
+    this.worklet.port.postMessage({ type: "clearInterrupted" });
```
kraenhansen, line 178: This used to be guarded behind a check:

https://github.com/elevenlabs/packages/blob/d28e2a9539c6742042fbe12d3e44778eb1b810dd/packages/client/src/VoiceConversation.ts#L171

I'm not 💯 on what the purpose of that was and if we need this still? The way I read it, we were only doing this if the new audio is after an active interruption.

kraenhansen (approved):
If it works I see no reason to hold it back 👍

packages/client/src/utils/output.ts
Hashir-Majeed: we keep the guard in VoiceConversation `handleAudio` method

Hashir-Majeed: I don't think after the refactor the new method has knowledge of event id anyways, so I think it's ok

Hashir-Majeed: tested it out with tool call sounds which notably manipulate event id, and it worked fine

PR #764, opened 2026-05-09
build: implement cross-platform version generation

gaureshpai:
## Summary

This PR fixes several build and tooling issues that prevented the workspace from building reliably on native Windows environments (without WSL), while also simplifying task orchestration in Turbo.

## Key Changes

* Replaced shell-specific `printf` commands with a cross-platform Node.js utility script (`scripts/generate-version.mjs`) to generate `version.ts` files consistently across packages.
* Updated package scripts to use `node --run` for internal task execution, improving consistency and reliability.
* Refined TypeScript build commands to explicitly use `tsconfig.build.json` where needed, preventing test files from being included in production builds and avoiding `TS6305` / inference errors.
* Standardized Prettier glob patterns to use double quotes, ensuring correct expansion on Windows shells.
* Updated `@elevenlabs/client` worklet generation to automatically format generated TypeScript artifacts, preventing lint regressions.
* Added a root `generate` Turbo task and simplified task dependencies by relying on transitive `generate` dependencies instead of repeating individual `generate-*` tasks across `build`, `dev`, and `check-types`.

## Packages Affected

* `@elevenlabs/client`
* `@elevenlabs/react`
* `@elevenlabs/react-native`
* `@elevenlabs/convai-widget-core`
* `@elevenlabs/convai-widget-embed`
* `@elevenlabs/types`

## Verification

* `pnpm build` succeeds across the full workspace
* `pnpm lint` passes
* `pnpm check-types` passes across packages and examples
* Verified on native Windows (without WSL)

---

> [!NOTE]
> **Low Risk**
> Changes are limited to build scripts, generated artifacts, and lint/format configuration with no product runtime or security logic.
> 
> **Overview**
> Replaces per-package shell `printf` **generate-version** steps with a shared Node script (`scripts/generate-version.mjs`) so `src/version.ts` is written the same way on Windows and Unix. Several packages now call that script; Turbo’s **generate-version** task also depends on the script as an input.
> 
> Adds a root **`pnpm generate`** script (`turbo run generate`) and tightens lint/format tooling: Prettier globs use double quotes for reliable expansion on Windows, **convai-widget-embed**’s prettier lint runs with **`--check`**, and **@elevenlabs/client** formats generated web worklet TypeScript after generation. **@elevenlabs/react** drops a **`prebuild`** hook that only ran version generation (handled via Turbo **generate** instead).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f429092c9866184fbb9e1cde077024a50c79edb0. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


scripts/generate-version.mjs
```diff
@@ -0,0 +1,23 @@
+import { writeFileSync, mkdirSync } from 'fs';
+import { dirname } from 'path';
```
kraenhansen:
```suggestion
import { writeFileSync, mkdirSync } from 'node:fs';
import { dirname } from 'node:path';
```

kraenhansen (approved):
Thanks!

I'm happy to take this exact PR - been wanting to turn the printfs into a proper script for some time now. I'd prefer the generate-version to be typescript, but it's not a blocker for merging (as we already have other mjs script files in the repo).

More generally speaking, we're developing on macs internally and can't guarantee this won't regress in the future.

gaureshpai:
Thanks for the review! I've verified the script builds correctly on Windows. Would appreciate a test on macOS to ensure no issues there.

packages/react-native/package.json
```diff
@@ -11,13 +11,13 @@
     }
   },
   "scripts": {
-    "generate-version": "printf \"// This file is auto-generated during build\\nexport const PACKAGE_VERSION = \\\"%s\\\";\\n\" \"$npm_package_version\" > src/version.ts",
-    "build": "tsc --build tsconfig.build.json",
+    "generate-version": "node ../../scripts/generate-version.mjs",
+    "build": "node --run generate-version && tsc --build tsconfig.build.json",
```
kraenhansen: I'd expect turbo to handle generating before building? I don't think we need this addition 🤔

gaureshpai: fair, will do :)

gaureshpai: Done :)

kraenhansen: Not what I meant. Are you aware this is a turbo repo mono-repository? We should use the turbo tasks and their declared dependencies to drive generation before building.

gaureshpai: I am aware :)

I got confused here 

> I'd expect turbo to handle generating before building?

kraenhansen: Okay. To clarify, I don't think we should have a "prebuild" script nor inlining a call to the generate script, I'd like each of these package scripts to be a single command and have turbo handle their interdependencies.

gaureshpai: Got it! thank you

gaureshpai: Did the required changes, please correct me if it is wrong, thank you so much

turbo.json
```diff
@@ -32,24 +43,39 @@
...
-      "dependsOn": ["generate", "^build"]
+      "dependsOn": [
+        "generate-version",
+        "generate-worklets",
+        "generate-types",
+        "^build"
+      ]
     },
     "generate": {
       "dependsOn": ["generate-types", "generate-version", "generate-worklets"]
```
kraenhansen, line 55: I don't think you need to many changes to this file? See how `generate` depends on these more specialised `generate-*` scripts.

kraenhansen (changes requested):

turbo.json
```diff
@@ -4,7 +4,12 @@
   "globalEnv": ["NODE_ENV"],
   "tasks": {
     "build": {
-      "dependsOn": ["generate", "^build"],
+      "dependsOn": [
+        "generate-version",
+        "generate-worklets",
+        "generate-types",
```
kraenhansen: This is already transitively covered by the dependencies of the `generate` task.

gaureshpai: Updated!

packages/client/package.json
```diff
@@ -13,15 +13,15 @@
     "./internal": "./dist/internal.js"
   },
   "scripts": {
-    "generate-version": "printf \"// This file is auto-generated during build\\nexport const PACKAGE_VERSION = \\\"%s\\\";\\n\" \"$npm_package_version\" > src/version.ts",
-    "generate-worklets": "node scripts/generateWorklets.js",
-    "build:esm": "tsc --build",
+    "generate-version": "node ../../scripts/generate-version.mjs",
+    "generate-worklets": "node scripts/generateWorklets.js && prettier src/utils/*.generated.ts --write",
```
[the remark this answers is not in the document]
kraenhansen: @gaureshpai in general I like the addition to invoke prettier here, but also this seems valuable to fix to keep consistency, by adding double quotes, to let Prettier expand these paths. There's also the option to extend the generator script to apply prettier before / after the write to the file 🤔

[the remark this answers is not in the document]
gaureshpai: added double quotes with escape characters \

gaureshpai:
@kraenhansen Should I change anything please?

kraenhansen (approved):
Seems correct to me. Thanks! And sorry for the wait on my part here 🙏

gaureshpai:
> Seems correct to me. Thanks! And sorry for the wait on my part here 🙏

No worries. Appreciate it :)

PR #660, opened 2026-04-09
[Widget] Use server side message ordering for consistency

kalvin807:
## Summary
- In voice mode, the server can send `agent_response` before `user_transcript` (agent starts generating while speech is still being transcribed)
- The widget appended messages in arrival order, causing agent responses to appear before the user's message
- Fix: sort display transcript entries by server-assigned `eventId` to ensure correct conversational ordering

Closes AP-1747

## Test plan
- [x] Added unit tests for eventId sorting in `buildDisplayTranscript`
- [x] Existing tests pass (24/24)
- [ ] Manual test: voice conversation should show user message before agent response

---

> [!NOTE]
> **Medium Risk**
> Changes transcript ordering logic, which could affect how mixed message/non-message entries render and how tool statuses attach if eventIds are missing or duplicated. Scope is limited and covered by new unit tests for multiple ordering scenarios.
> 
> **Overview**
> Fixes voice-mode transcript ordering by **re-sorting message entries with an `eventId`** in `buildDisplayTranscript`, ensuring user transcripts appear before agent responses when events arrive out of order, while leaving non-`eventId` entries in place.
> 
> Adds targeted unit tests for `eventId` sorting behavior (including mixed non-message entries) and ships a patch changeset for `@elevenlabs/convai-widget-core`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 5883c8099a70ded7caa3220432ceca0839730d8c. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #767, opened 2026-05-10
[Widget] Fix same-turn user/agent transcript ordering race

kalvin807:
## Why

In voice/DTMF mode the agent message can render before the user transcript. Cause is asymmetric send dispatch in `xi/services/python/xi_backend/xi_backend/routes/convai/utils/chat_history.py:1089-1172`: `user_transcript` (audio + DTMF path) is sent via `asyncio.create_task(send_json(...))` while `agent_response` is sent via `await send_json(...)`. The agent send can win the wire.

Both events for the same turn carry the same `source_event_id` (chat_history.py:1103-1106 and 1167-1170), so they reach the widget with an identical `eventId`. PR #660's sort had no tiebreaker on equal IDs, so display order tracked arrival order.

## What

- Sort comparator in `display-transcript.ts` now uses `(eventId, role[user-first], originalIndex)`.
- Rows without `eventId` (local optimistic user echoes from `sendUserMessage`) keep their original slots; non-message rows stay fixed.
- Tool-status attachment runs after the reorder; keys off `eventId` so it is unaffected.

## Testing

Added regression cases in `display-transcript.test.ts`:
- same `eventId` with agent received first → user displays first
- agent-only turn (no synthetic user inserted)
- local user message without `eventId` stays anchored
- non-message rows between misordered messages stay in place

Manual test

---

> [!NOTE]
> **Medium Risk**
> Adjusts transcript sorting logic in `buildDisplayTranscript`, which can subtly change message ordering in voice/DTMF flows and could regress edge-case rendering if the comparator is wrong. Coverage is improved with new regression tests, reducing risk.
> 
> **Overview**
> Fixes a race where voice/DTMF turns could display the agent message before the user transcript when both arrive with the same `eventId`.
> 
> `buildDisplayTranscript` now sorts server-stamped message rows by `(eventId, user-before-agent, originalIndex)` while keeping non-`eventId` messages and non-message rows anchored in place; new tests cover same-`eventId` tie-breaking and related ordering edge cases. Includes a patch `changeset` for `@elevenlabs/convai-widget-core`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit af19518cf4c1cd8cd94f5b6381a79d81c3b26009. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/fix-widget-cross-conversation-ordering.md
+++ b/.changeset/fix-widget-cross-conversation-ordering.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/convai-widget-core": patch
+---
+
+Scope transcript reordering by `conversationIndex` and break ties on shared `eventId` with user-before-agent. Fixes two issues in voice/DTMF mode: (1) cross-conversation interleaving when `event_id` resets between sessions, and (2) agent message rendering before user transcript when both share the same turn `event_id`.
--- a/.changeset/fix-widget-cross-conversation-ordering.md
+++ b/.changeset/fix-widget-cross-conversation-ordering.md
@@ -2,4 +2,4 @@
 "@elevenlabs/convai-widget-core": patch
 ---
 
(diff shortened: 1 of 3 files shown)


kamilk-11 (approved):
LGTM

PR #775, opened 2026-05-11
[Widget] Handle voice transcript ordering on insert

kalvin807:
Why:
Voice sessions can receive final agent responses before matching user transcripts, while also receiving agent response parts. The widget should avoid treating voice response parts as transcript messages and should keep same-turn user transcripts before their agent response when the server transcript arrives late.

What:
- Ignore `agent_chat_response_part` transcript updates for voice sessions while preserving text-only streaming behavior.
- Insert late server user transcripts before an existing matching agent message for the same conversation and event id.
- Remove broad display-time event id sorting so `buildDisplayTranscript` only filters, groups, and attaches tool status.

Testing:
- Updated `display-transcript` unit coverage for source-order preserving display behavior.

---

> [!NOTE]
> **Medium Risk**
> Changes transcript ordering logic for live conversations and removes display-time eventId sorting, which could alter message sequencing in edge cases across voice/text sessions. Impact is user-visible but limited to transcript rendering/stream handling.
> 
> **Overview**
> Fixes voice-session transcript behavior by **ignoring streamed `agent_chat_response_part` updates unless in text-only mode** and by **inserting late-arriving server user transcripts before the matching agent message** for the same `eventId`/`conversationIndex`.
> 
> Removes broad eventId-based reordering from `buildDisplayTranscript`, so display building now preserves source order (while still filtering/grouping and attaching tool status), and updates unit tests accordingly.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit bce369428045d3a6180703e4e10046eacc09f254. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/convai-widget-core/src/contexts/conversation.tsx
+++ b/packages/convai-widget-core/src/contexts/conversation.tsx
@@ -76,6 +76,35 @@ export type TranscriptEntry =
       conversationIndex: number;
     };
 
+// Voice user transcripts can arrive after the agent response for the same
+// server turn. Insert those late user messages before the matching agent row;
+// keep local text/multimodal messages anchored by normal append order.
+function appendTranscriptMessage(
+  entries: TranscriptEntry[],
+  entry: Extract<TranscriptEntry, { type: "message" }>
+): TranscriptEntry[] {
+  if (entry.role !== "user" || entry.eventId == null) {
+    return [...entries, entry];
+  }
+
+  const matchingAgentIndex = entries.findIndex(
+    candidate =>
+      candidate.type === "message" &&
+      candidate.role === "agent" &&
+      candidate.eventId === entry.eventId &&
+      candidate.conversationIndex === entry.conversationIndex
+  );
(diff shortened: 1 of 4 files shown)


packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -252,22 +281,22 @@ function useConversationSetup() {
...
+                type: "message",
+                role,
+                message,
+                isText: false,
+                conversationIndex: conversationIndex.peek(),
+                eventId: event_id,
+              });
             },
             onAgentChatResponsePart: ({ text, type, event_id }) => {
+              if (conversationTextOnly.peek() !== true) {
```
kalvin807, line 294: 📝 in voice mode, we should ignore the partial. to make ordering more stable. and it is flaky in speculative turn

kamilk-11 (approved):
LGTM

PR #793, opened 2026-05-19
Add new environment for publish action

PaulAsjes:
Require approvals for releasing new versions of each package.

---

> [!NOTE]
> **Medium Risk**
> Low code complexity, but it changes the release/publishing pipeline and could block or delay package publication if the `npm-publish` environment is misconfigured.
> 
> **Overview**
> Adds `environment: npm-publish` to the `Publish package` GitHub Actions workflow, enabling environment-based protections (e.g., required reviewers) to gate npm publishing on release events.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a5dbf8726d5066ff8ba0e4c7687fb53e1c78b5c2. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

PR #807, opened 2026-05-20
Delete `.github/workflows/publish.yml`

kraenhansen:
Follow-up to https://github.com/elevenlabs/packages/pull/793 I think we should just delete this now that we're using Changesets for releases.

---

> [!NOTE]
> **Low Risk**
> Low risk: deletes an unused GitHub Actions workflow and does not change runtime or package code. Only affects release/publish automation if any consumers still relied on the removed `publish.yml`.
> 
> **Overview**
> Removes the legacy `.github/workflows/publish.yml` workflow that published packages on `release.published`, including version/tag validation and workspace dependency checks.
> 
> Publishing remains driven by the existing Changesets-based `release.yml` workflow (`changesets/action` calling `pnpm run publish`).
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit ab76250c8f9b5709eaa4ceec545278cbc8e9e83a. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


.github/workflows/publish.yml
```diff
@@ -1,141 +0,0 @@
-name: Publish package
-
-on:
-  release:
-    types: [published]
-
-jobs:
-  publish-npm:
-    runs-on: ubuntu-latest
-    environment: npm-publish
```
PaulAsjes, line 10: Hmm is this not used as part of the flow? If changesets does the publishing then we need to update it to use the right environment.

kraenhansen: It's not used at all for publishing no - just kept it around in case we needed to revert the changeset stuff, but that seems overdue by now 🙂

PaulAsjes (approved):

PR #817, opened 2026-05-26
Configurable audio input chunk length and default 25ms instead of 100ms

ksergazy:
Lower raw audio worklet buffer threshold to 25ms so mic chunks are sent to the agent more frequently for lower latency.

---

> [!NOTE]
> **Medium Risk**
> Changes real-time uplink buffering and message frequency on the WebSocket voice path; misconfigured values could affect latency or server load, but defaults are conservative and the option is create-time only.
> 
> **Overview**
> Adds **`inputChunkDurationMs`** on the WebSocket microphone path so callers can control how long each mic chunk is before it is sent to the agent (default **25ms**, exported as `DEFAULT_INPUT_CHUNK_DURATION_MS`).
> 
> The **raw audio AudioWorklet** no longer uses a fixed `sampleRate / 10` (~100ms) buffer; it sets `bufferSize` from `chunkDurationMs` (with a minimum of one sample). **`MediaDeviceInput`** and **`VoiceSessionSetup`** pass the option through on session setup; it cannot be changed after create without recreating the input.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit b6c01886741c20cb5598642670ec8a984b2627ba. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/reduce-audio-chunk-length-10ms.md
+++ b/.changeset/reduce-audio-chunk-length-10ms.md
@@ -1,6 +1,6 @@
 ---
-"@elevenlabs/client": patch
+"@elevenlabs/client": minor
 "@elevenlabs/react": patch
 ---
 
-Reduce audio input chunk length from 100ms to 10ms for lower latency.
+Make microphone input chunk duration configurable via `inputChunkDurationMs` (default 25ms).
--- a/.changeset/reduce-audio-chunk-length-10ms.md
+++ b/.changeset/reduce-audio-chunk-length-10ms.md
@@ -1,0 +1,6 @@
+---
+"@elevenlabs/client": patch
+"@elevenlabs/react": patch
+---
+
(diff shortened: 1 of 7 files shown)


kraenhansen:
Did we test the impact of this? What's the cost and how did we decide this was the right tradeoff? I'm wondering if we'd want to make this configurable - perhaps just internally facing at first.

ksergazy:
Majority of API-level clients already send lower sub 10ms chunks. Making it configurable as suggested but with default of 25ms, it should be safe.

AngeloGiacco (approved):
it looks fine to me please QA

AngeloGiacco:
and please verify xi is setup for it like jozef mentioned

PR #827, opened 2026-06-03
Lucassonnabend/support takeover typing indicators

lucas-sonnabend:
Add typing indicator when agent is typing during externalAgent mode.

I realised that this doesn't have a typing indicator at all, so I added one via 3 simple dots.

3Shape is asking about this for their genesys integration, and it makes sense, because seeing the external agent typing makes it feel a lot more responsive.

Loom: https://www.loom.com/share/800f1020a4114af9bdfc800019e72bfa

---

> [!NOTE]
> **Low Risk**
> Mostly new optional WebSocket events and transcript UI; no auth, billing, or core session security changes.
> 
> **Overview**
> Adds **external-agent takeover UX**: the client and widget now handle **`agent_typing`** and **`external_agent_connected`** WebSocket events (AsyncAPI, generated types, `BaseConversation`, React hook callbacks).
> 
> When an external agent connects, the widget enters external-agent mode and shows a **shimmering “Agent is typing …”** line at the end of the transcript while `is_typing` is true (optional `duration_ms` auto-clear). Typing clears on agent messages, streaming parts, and disconnect. **`StatusLabel`** updates speaking/listening labels immediately when the agent is speaking.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e7824e359d16a6903e1c0b0ba960a2d36fb2ecbf. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


AngeloGiacco:
very very very cool work

AngeloGiacco:
can we make it only disappear if there is no text? that would feel more natural to me

AngeloGiacco (approved):
lgtm lets consider removing the 5second timeout in the case of text filled in

packages/convai-widget-core/src/widget/StatusLabel.tsx
```diff
@@ -7,32 +7,38 @@ import { useTextContents } from "../contexts/text-contents";
...
-  className,
-  ...props
-}: HTMLAttributes<HTMLDivElement>) {
+
+function userCurrentLabel() {
   const { status, isSpeaking } = useConversation();
   const textOnly = useIsConversationTextOnly();
   const { isTextMode } = useConversationMode();
   const text = useTextContents();
-  const currentLabel = useComputed(() =>
```
kalvin807, line 18: 🔥 nice cleanup

packages/convai-widget-core/src/styles/index.css
```diff
@@ -106,6 +102,34 @@
     animation: spin 1s linear infinite;
   }
 
+  @keyframes my-bounce {
```
kalvin807: how about `typing-dot-bounce`

kalvin807 (approved):
lgtm. 
btw did you consider using this indicator, https://ui.elevenlabs.io/docs/components/shimmering-text

it is used by our in-app agent when agent is thinking (which similar to typing), imo reusing for consistency 

also ping me if you need a release

lucas-sonnabend:
> btw did you consider using this indicator, https://ui.elevenlabs.io/docs/components/shimmering-text
> 
> it is used by our in-app agent when agent is thinking (which similar to typing), imo reusing for consistency

Great idea, I actually had something similar in mind, but I didn't know we already had a component for it, and then the 3 dots were simpler to implement.
I'll use that instead.

lucas-sonnabend:
> can we make it only disappear if there is no text? that would feel more natural to me

At the moment we only get "isTyping true/false" events from genesys, we have no way of knowing if there is still a draft or not, so the best we can do is to have it disappear after a timeout.
I think this also matches other typing indicators like slack, for example

lucas-sonnabend:
updated with shimmering text:
https://www.loom.com/share/e0b67f6c32a545e48bfcfb88c1625348

PR #848, opened 2026-06-17
feat(client): add ASR keywords override to session config

amandamilberg:
## Summary
- Add `overrides.asr.keywords` to `@elevenlabs/client` `SessionConfig` so browser SDK users can send per-conversation ASR keyword biasing via `conversation_initiation_client_data`.
- Extend the AsyncAPI schema and generated `@elevenlabs/types` `ConversationConfigOverride` with `asr.keywords`.
- Forward the override in `constructOverrides()` and add unit tests.

## Usage
```ts
await Conversation.startSession({
  agentId: "your-agent-id",
  overrides: {
    asr: {
      keywords: ["Cvent", "registration"],
    },
  },
});
```

**Note:** The agent must permit `asr.keywords` overrides in its security/override settings (off by default). The backend silently ignores this override when not permitted.

## Test plan
- [x] `pnpm exec vitest run src/utils/overrides.test.ts --browser.headless` in `packages/client`
- [x] Pre-commit turbo lint/build passed locally
- [ ] Verify with an agent that has `asr.keywords` override enabled

Made with [Cursor](https://cursor.com)

### DIFF
--- a/packages/client/src/utils/overrides.ts
+++ b/packages/client/src/utils/overrides.ts
@@ -25,9 +25,9 @@ export function constructOverrides(
         stability: config.overrides.tts?.stability,
         similarity_boost: config.overrides.tts?.similarityBoost,
       },
-      asr: {
-        keywords: config.overrides.asr?.keywords,
-      },
+      ...(config.overrides.asr?.keywords !== undefined
+        ? { asr: { keywords: config.overrides.asr.keywords } }
+        : {}),
       conversation: {
         text_only: config.overrides.conversation?.textOnly,
       },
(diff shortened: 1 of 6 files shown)


kraenhansen (commented):
@amandamilberg it looks like the tests just needs a small update to pass - otherwise the code looks good 👍

packages/client/src/utils/overrides.ts
```diff
@@ -25,6 +25,9 @@ export function constructOverrides(
         stability: config.overrides.tts?.stability,
         similarity_boost: config.overrides.tts?.similarityBoost,
       },
+      ...(config.overrides.asr?.keywords !== undefined
+        ? { asr: { keywords: config.overrides.asr.keywords } }
+        : {}),
```
kraenhansen, line 30: This is growing a bit complex, I feel its ripe for a refactor (something which strips nested objects with no property values from the object returned from this function) - if not in this PR, its a potential followup for you or someone else in the future 🙂

kraenhansen (approved):

PR #851, opened 2026-06-19
[Widget] Re-enable text streaming in voice sessions

bsreenivas11:
## Problem
PR #775 disabled `agent_chat_response_part` streaming for voice sessions because the backend was emitting these events eagerly from abandoned turns, making them unreliable. It also added an ordering hack in `appendTranscriptMessage` to insert late user transcripts before the matching agent response, because user transcripts could arrive after the agent response for the same turn.

## Fix
Two backend fixes have since removed the root causes:
- elevenlabs/xi#37097 gates `agent_chat_response_part` events on `turn_irreversible`, so they are only emitted once the turn is committed. The voice exclusion is no longer needed.
- elevenlabs/xi#38653 sends the user transcript before opening the gate, guaranteeing user transcripts always arrive before response parts. The ordering hack is no longer needed.

This restores the original behavior: the transcript streams in word-by-word as the agent responds in both voice and text-only sessions, with messages in the correct order.

## Changes
- Remove the early return in `onAgentChatResponsePart` that skipped all voice session events
- Restore the `conversationTextOnly` guard on the `firstMessage` skip (text-only-specific behavior that was previously made redundant by the early return)
- Remove the ordering hack in `appendTranscriptMessage` and inline the plain append at the call site (it can never trigger now that user transcripts arrive first)

## Test plan
[DONE] Local widget connected to prod server, voice + transcript enabled — verified user messages always appear above their matching agent response across multiple turns, and agent text streams in word-by-word.

## Dependency
[DONE] Requires elevenlabs/xi#37097 and elevenlabs/xi#38653 to be deployed (both merged), otherwise voice sessions may display partial text from abandoned turns or out-of-order messages.

## Why this is safe
Bugbot may flag the removal of the ordering logic in `appendTranscriptMessage` as a potential out-of-order risk. It is safe because the race it defended against no longer exists:

- The hack only triggered when a user transcript arrived **after** the matching agent response for the same turn.
- elevenlabs/xi#38653 moves `turn.turn_irreversible.set()` to run **after** `add_user_message`, so the backend now sends the user transcript before opening the gate that allows any `agent_chat_response_part` events through.
- Therefore, by the time agent response parts (or the final agent message) reach the client, the matching user transcript is already in the transcript. The `matchingAgentIndex` lookup would always return -1, i.e. the hack was already a no-op against the fixed backend.

Verified on prod (voice + transcript): user messages always render above their matching agent response across multiple turns.

---

> [!NOTE]
> **Medium Risk**
> Transcript order and partial text in voice depend on deployed backend turn-ordering fixes; wrong deployment could regress UX but does not touch auth or data handling.
> 
> **Overview**
> **Re-enables word-by-word agent transcript streaming in voice sessions** by processing `agent_chat_response_part` events again instead of ignoring them when the session is not text-only.
> 
> The **`appendTranscriptMessage` ordering helper is removed**; non-streaming messages are appended in arrival order, relying on backend guarantees that user transcripts arrive before agent response parts. **Text-only-only behavior is restored** for skipping the first agent stream when `firstMessage` is set, and streamed/final agent rows set **`isText` from `conversationTextOnly`** rather than always treating streams as text.
> 
> Patch changeset notes the streaming fix for `@elevenlabs/convai-widget-core`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit e1045aeae09c5360c500f110e54266762ddd1c48. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/convai-widget-core/src/contexts/conversation.tsx
+++ b/packages/convai-widget-core/src/contexts/conversation.tsx
@@ -280,7 +280,7 @@ function useConversationSetup() {
                     type: "message",
                     role: "agent",
                     message,
-                    isText: true,
+                    isText: conversationTextOnly.peek() === true,
                     conversationIndex: conversationIndex.peek(),
                     eventId: event_id,
                   };
@@ -325,7 +325,7 @@ function useConversationSetup() {
                     type: "message",
                     role: "agent",
                     message: "",
-                    isText: true,
+                    isText: conversationTextOnly.peek() === true,
                     conversationIndex: conversationIndex.peek(),
                     eventId: event_id,
                   },
--- a/packages/convai-widget-core/src/contexts/conversation.tsx
+++ b/packages/convai-widget-core/src/contexts/conversation.tsx
@@ -81,33 +81,11 @@ export type TranscriptEntry =
       conversationIndex: number;
     };
 
-// Voice user transcripts can arrive after the agent response for the same
-// server turn. Insert those late user messages before the matching agent row;
-// keep local text/multimodal messages anchored by normal append order.
 function appendTranscriptMessage(
   entries: TranscriptEntry[],
   entry: Extract<TranscriptEntry, { type: "message" }>
 ): TranscriptEntry[] {
-  if (entry.role !== "user" || entry.eventId == null) {
-    return [...entries, entry];
-  }
-
-  const matchingAgentIndex = entries.findIndex(
-    candidate =>
-      candidate.type === "message" &&
-      candidate.role === "agent" &&
-      candidate.eventId === entry.eventId &&
-      candidate.conversationIndex === entry.conversationIndex
-  );
-  if (matchingAgentIndex === -1) {
-    return [...entries, entry];
-  }
-
-  return [
-    ...entries.slice(0, matchingAgentIndex),
-    entry,
-    ...entries.slice(matchingAgentIndex),
-  ];
+  return [...entries, entry];
 }
 
 export function ConversationProvider({ children }: ConversationProviderProps) {
--- a/packages/convai-widget-core/src/contexts/conversation.tsx
+++ b/packages/convai-widget-core/src/contexts/conversation.tsx
@@ -81,13 +81,6 @@ export type TranscriptEntry =
       conversationIndex: number;
     };
 
-function appendTranscriptMessage(
-  entries: TranscriptEntry[],
-  entry: Extract<TranscriptEntry, { type: "message" }>
-): TranscriptEntry[] {
-  return [...entries, entry];
-}
-
 export function ConversationProvider({ children }: ConversationProviderProps) {
   const value = useConversationSetup();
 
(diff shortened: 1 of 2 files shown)


renal128 (commented):
lgtm! were you able to test it? also would be great to get @kalvin807 's eyes on this

kalvin807 (approved):
Tested locally, lgtm. 
please add a changeset describe what changed and also trigger ci.
Question:
1. Do post-fix backend can guarantee position so that we can remove ordering hack at `function appendTranscriptMessage` as well?

renal128:
> 1. Do post-fix backend can guarantee position so that we can remove ordering hack at `function appendTranscriptMessage` as well?

@kalvin807 good question! I think this PR should fix the order between full response and request messages, but not quite sure about response parts: https://github.com/elevenlabs/xi/pull/35526

if you look at the current code here: https://github.com/elevenlabs/xi/blob/main/services/python/xi_backend/xi_backend/routes/convai/utils/chat_history.py#L1257-L1263

```
        turn.turn_irreversible.set()
        turn.previous_turn_cancelled_due_to_asr_change = False
        for i, message in enumerate(turn.turn_chat):
            if message.role != ChatRole.USER:
                # Expect user transcripts to appear only at the beginning
                break
            await self.add_user_message(message)
```
  

- `turn.turn_irreversible.set()` is what opens the gate for sending response parts 
- `await self.add_user_message(message)` sends user transcript

so there might be a race there between response parts and user transcript... ~@bsreenivas11 would you mind making the change to move `turn.turn_irreversible.set()` to be after `await self.add_user_message(message)`? apologies for missing this in the gating PR~
[upd] actually let me do that real quick: https://github.com/elevenlabs/xi/pull/38653
[upd 2] merged ^

packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -329,11 +329,11 @@ function useConversationSetup() {
...
-              if (conversationTextOnly.peek() !== true) {
-                return;
-              }
-
-              if (firstMessage.peek() && !receivedFirstMessageRef.current) {
+              if (
+                firstMessage.peek() &&
+                conversationTextOnly.peek() === true &&
+                !receivedFirstMessageRef.current
+              ) {
```
kamilk-11, line 310: @kalvin807 IIRC you mentioned that ignoring streaming fixed an issue with invalid transcript ordering? Just making sure if this is now solved or will reverting this change bring the issue back too?

kamilk-11: Nvm, saw the discussion above now

kamilk-11 (approved):

packages/convai-widget-core/src/contexts/conversation.tsx
kalvin807: `appendTranscriptMessage` fn will still sort the message so ordering should be all good.
And @renal128 just made this 
https://github.com/elevenlabs/xi/pull/38653 which gonna fix the root cause.

renal128: merged my change

packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -81,33 +81,11 @@ export type TranscriptEntry =
...
-  if (matchingAgentIndex === -1) {
-    return [...entries, entry];
-  }
-
-  return [
-    ...entries.slice(0, matchingAgentIndex),
-    entry,
-    ...entries.slice(matchingAgentIndex),
-  ];
+  return [...entries, entry];
```
kalvin807: Nit: we should just inline this logic, now this function basically do nothing 
Tip: You can update the pr description above why this change is safe so bugbot no longer complains about it again.

kalvin807 (approved):
Before merge, please connect local widget to prod server to check if ordering really works. It only happens with voice + transcript

bsreenivas11:
image

Ordering works with user turn shown before agent response

PR #856, opened 2026-06-24
[agents][feedback] allow sendFeedback to target a past message by event_id

AngeloGiacco:
this is compliant with our asyncapi spec and backend supports it

this would unblock some work of mine in the xi repo that consumes the client and wants to send the event_id

### testing 
added a test

---

> [!NOTE]
> **Medium Risk**
> Public API and semantics of `canSendFeedback` change (always true when connected vs one-shot per turn); apps that hid feedback after the first like/dislike need to adjust UX, though behavior matches the async API and backend.
> 
> **Overview**
> **`sendFeedback`** now accepts an optional **`eventId`**: omit it to rate the latest agent turn (tracked via **`currentEventId`** on each **`agent_response`**), or pass an id to rate a specific past message. The wire payload still uses **`event_id`** on the feedback message.
> 
> **`canSendFeedback`** no longer blocks after one rating per turn. It is **`true` while the session is `connected`** (updated on status changes), so callers can rate any message or re-rate during a live conversation. One-feedback-per-turn tracking (**`lastFeedbackEventId`**) was removed.
> 
> **`@elevenlabs/react`** **`useConversationFeedback`** forwards the optional **`eventId`** to the client. Minor version bumps are noted in the changeset.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a58b3019b75693d2edc2055b0b36e84b84abd3a1. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/BaseConversation.ts
+++ b/packages/client/src/BaseConversation.ts
@@ -118,7 +118,6 @@ export abstract class BaseConversation {
   protected status: Status = "connecting";
   protected volume = 1;
   protected currentEventId = 1;
-  protected lastFeedbackEventId = 0;
   protected canSendFeedback = false;
 
   protected static getFullOptions(partialOptions: PartialOptions): Options {
@@ -196,11 +195,12 @@ export abstract class BaseConversation {
       if (this.options.onStatusChange) {
         this.options.onStatusChange({ status });
       }
+      this.updateCanSendFeedback();
     }
   }
 
   protected updateCanSendFeedback() {
-    const canSendFeedback = this.currentEventId !== this.lastFeedbackEventId;
+    const canSendFeedback = this.status === "connected";
     if (this.canSendFeedback !== canSendFeedback) {
       this.canSendFeedback = canSendFeedback;
       if (this.options.onCanSendFeedbackChange) {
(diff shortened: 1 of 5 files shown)


packages/client/src/BaseConversation.ts
```diff
@@ -593,23 +593,17 @@ export abstract class BaseConversation {
...
-      );
+      console.warn("Cannot send feedback: the conversation is not connected.");
       return;
     }
 
     this.connection.sendMessage({
       type: "feedback",
       score: like ? "like" : "dislike",
-      event_id: this.currentEventId,
+      event_id: eventId ?? this.currentEventId,
```
renal128, line 606: is it possible for `this.currentEventId` to be empty here?

AngeloGiacco: nope

renal128: eh, cursor was faster haha ^

renal128 (approved):
lgtm

kraenhansen (commented):
LGTM

PR #861, opened 2026-06-26
[Widget] Don't merge agent text segments split by a tool call

bsreenivas11:
## Summary
Fixes a bug where two agent text segments in the same turn, separated by a tool call (text → tool → text), would collapse into one — the second text bubble overwrote the first in the rendered transcript.

## Root cause
`buildDisplayTranscript` merges consecutive messages that share an `eventId` + role, to collapse a streamed partial into its finalized version. But tool entries are skipped during result building, so two distinct agent text segments separated by a tool call appear "consecutive" — and since all segments in a turn share one `eventId`, the second segment overwrote the first.

This is a pre-existing bug from #528 (2026-02-24). It became visible once `agent_chat_response_part` streaming was re-enabled for voice sessions, which routes voice agent text through `eventId`-tagged rows.

## Fix
Track whether a tool entry appeared since the last emitted message (`toolBetween`). Skip the same-`eventId` merge when a tool call separated the two messages — those are distinct bubbles, not a partial/final pair. The streamed-partial/final collapse is unchanged.

## Test plan
- [x] New unit tests in `display-transcript.test.ts`:
  - text → tool → text segments stay as two separate bubbles
  - a streamed partial/final pair (no tool between) still collapses to the final message

BEFORE:

https://github.com/user-attachments/assets/e109f00d-f9bb-42ef-bb95-1cbc1c9eecd1

AFTER:

https://github.com/user-attachments/assets/36ef1212-4767-4ddc-b5c1-824ae012795a

---

> [!NOTE]
> **Low Risk**
> Scoped to convai widget transcript display logic with added unit tests; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes **transcript rendering** when a single agent turn interleaves **text → tool → text** with the same `eventId`. Those segments no longer collapse into one bubble (where the later text overwrote the earlier).
> 
> `buildDisplayTranscript` tracks whether a **tool request/response** sat between consecutive messages (`toolBetween`). Same-`eventId` merges still collapse streamed partials and empty placeholders, but merges are skipped when a tool separated two **non-empty** text segments. Empty placeholders before/after tools still fold into the final message.
> 
> With **agent tool status** enabled, the tool badge attaches to the **first** agent bubble for that turn only, so multi-bubble turns do not duplicate the status badge.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit a8baeba9491fe07e46b86cab62d19994a6ecbaab. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/tidy-tools-merge.md
+++ b/.changeset/tidy-tools-merge.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/convai-widget-core": patch
+---
+
+Fix transcript rendering for turns that interleave text and tool calls: keep text segments split by a tool call as separate bubbles, and show the tool status badge only once per turn
--- a/packages/convai-widget-core/src/utils/display-transcript.test.ts
+++ b/packages/convai-widget-core/src/utils/display-transcript.test.ts
@@ -283,6 +283,28 @@ describe("buildDisplayTranscript", () => {
       const result = build(input, { showAgentStatus: false });
       expect(result[0]).not.toHaveProperty("toolStatus");
     });
+
+    it("attaches status to only the first bubble of a text → tool → text turn", () => {
+      // Two non-empty segments share the turn's eventId and stay as separate
+      // bubbles. The tool status attaches to the first (triggering) bubble only,
+      // so the badge is not duplicated across both.
+      const input = [
+        msg("agent", "Running the tool now.", { eventId: 2 }),
+        toolReq(2),
+        toolRes(2),
+        msg("agent", "Tool completed successfully", { eventId: 2 }),
+      ];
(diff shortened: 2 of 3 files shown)


kalvin807 (approved):
Lgtm. Please add a changeset for release.

Question: does the bug occur if user do not enable this two tool related event?

bsreenivas11:
@kalvin807 the duplicate-badge symptom needs show-agent-status enabled, but the underlying text-loss bug does not — it happens for any text → tool → text turn because tool transcript entries are always created regardless of config.

PR #873, opened 2026-07-08
Release the microphone reliably on realtime teardown

### LINKED ISSUE #874: Scribe realtime (mic mode): microphone never released on close() race or server-initiated close
**Suggested fix:** #873 

---

**Bug 1 - close() during async mic setup.** 

`Scribe.connect({ microphone: {...} })`, then `connection.close()` shortly after the WebSocket opens (e.g. a press-and-hold UI where the user releases immediately). The socket `open` event fires `streamFromMicrophone()`, which awaits worklet load + `audioContext.resume()` before assigning `connection._audioCleanup`. If `close()` runs in that window, cleanup is skipped: the `MediaStreamTrack` and `AudioContext` are never torn down, the browser recording indicator stays lit, and the mic captures indefinitely. The worklet's `port.onmessage` then calls `connection.send()` per audio frame against the closed socket, throwing an uncaught `Error: WebSocket is not connected` on every frame.

---

**Bug 2 - server-initiated close.** 

The websocket `close` listener only emits events; `_audioCleanup` runs solely inside a user-called `close()`. When the server closes the socket (1006, quota exceeded, session time limit), a consumer that treats the CLOSE event as the end of the session is left with a live microphone and the same per-frame uncaught throws. No race needed.

---

**Minor:** the `error` and `close` listeners `console.error` and re-emit `ERROR` even when the close was app-initiated via `close()`. Closing a still-connecting socket always yields a 1006, so a deliberate quick teardown logs "WebSocket closed unexpectedly", which pollutes consoles and error trackers.

---

**Version:** `@elevenlabs/client` 1.14.1 (still present on `main`: `scribe.ts` `streamFromMicrophone`, `connection.ts` `close`/listeners)

keith-pw:
## Summary

Fixes a microphone leak and uncaught errors in Scribe realtime microphone mode when the connection is torn down. Closes #874 

`ScribeRealtime.streamFromMicrophone` assigns `connection._audioCleanup` only *after* an `await` (worklet load + `audioContext.resume()`). If the connection closes during that window, cleanup is skipped and the `MediaStreamTrack` + `AudioContext` are never released — the browser recording indicator stays lit and the mic keeps capturing. The now-orphaned worklet then calls `connection.send()` per audio frame against a closed socket, throwing an uncaught `WebSocket is not connected` on every frame. A server-initiated close hits the same leak, because `_audioCleanup` previously ran only inside a caller's `close()`.

This surfaces in any press-and-hold UI where the user releases quickly, and on any server-side disconnect (quota, session time limit, network 1006).

## Changes

- Track a `_closed` flag on `RealtimeConnection`, set by both `close()` and the socket `close` listener, so "the socket is gone" is known regardless of who closed it.
- Drop microphone frames once `_closed` instead of calling `send()` on a closed socket.
- Run microphone cleanup when the mic setup resolves after a close, and from the `close` listener itself, so **every** close path releases the mic. Both sites are guarded so cleanup never double-runs.
- Skip the "WebSocket closed unexpectedly" / `ERROR` emission when the app initiated `close()` — aborting a still-connecting socket yields a benign `1006` that was being logged as an error.

Public `send()` / `commit()` throw semantics are unchanged; only the SDK's own internal microphone caller is made defensive.

## Tests

Added coverage in `scribe.test.ts` using the existing `setScribeMicrophoneSetup` injection point and `mock-socket`:

- `close()` racing the async mic setup → cleanup runs, late frames are dropped and do not throw.
- server-initiated close → microphone is released even though the consumer never calls `close()`.
- `close()` aborting a still-connecting socket → no `ERROR` emitted, `CLOSE` still emitted.

All existing scribe tests continue to pass. Each new test was confirmed to fail against the unpatched source before the fix.

## Changeset

Included (`patch` bump for `@elevenlabs/client`).

---

> [!NOTE]
> **Medium Risk**
> Changes WebSocket close/error handling and mic lifecycle on all teardown paths; behavior shifts for early close and server disconnect, though scoped to Scribe realtime microphone mode.
> 
> **Overview**
> Fixes **microphone leaks** and spurious errors when Scribe realtime (microphone mode) shuts down before or without a full `close()` from the app.
> 
> `RealtimeConnection` now tracks **`_closed`** for both app `close()` and server socket close. Mic **`_audioCleanup`** runs from `close()`, from the socket **`close`** listener (server disconnect), and when async mic setup finishes after a close; **`_audioCleanup` is cleared** so cleanup does not double-run. **`streamFromMicrophone`** drops late audio frames when `_closed` instead of calling **`send()`** on a dead socket.
> 
> App-initiated **`close()`** no longer treats aborting a still-connecting socket as an unexpected **1006** (skips **`ERROR`** / noisy logs); real server-side abnormal closes still emit errors. Public **`send()`** / **`commit()`** throw behavior is unchanged.
> 
> Adds tests for close-vs-mic-setup race, server-initiated close without client `close()`, and early `close()` without **`ERROR`**. Patch changeset for **`@elevenlabs/client`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 35194e6040b4066b5d685de22aa817b2224ef7af. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


PaulAsjes (approved):
Looks good, thanks! Would you mind fixing up the conflicts before we merge?

keith-pw:
Hey @PaulAsjes Thank you for taking a look! I fixed the conflicts and pushed. The Lint/Test workflows are showing action_required - could you approve the workflow run so CI can complete? Thanks!

keith-pw:
@PaulAsjes conflicts resolved and CI is green - good to merge whenever you get a chance. TYSM!

PR #886, opened 2026-07-14
fix(client): guard handleErrorEvent against missing error_event payload

romka-best:
## Summary

`BaseConversation.handleErrorEvent` reads `event.error_event.error_type` (and `.message`, `.reason`, `.code`, `.debug_message`, `.details`) directly. When the orchestrator emits an `error` message whose `error_event` payload is absent/undefined, this throws:

```
TypeError: Cannot read properties of undefined (reading 'error_type')
```

The `"error"` branch of the `onMessage` dispatcher isn't wrapped in a try/catch (unlike `client_tool_call`), so the throw escapes `onMessage` and crashes the consumer's message/data-channel handler instead of being surfaced through `onError`. In a WebRTC (LiveKit) session this bubbles up out of the data-packet handler as an unhandled error and leaves the conversation wedged.

This is still present on `main` (`packages/client/src/BaseConversation.ts`).

## Change

- Read `event.error_event` once and access its fields with optional chaining, falling back to `"Unknown error"` for the message. Behavior for well-formed error events is unchanged; malformed ones now route to `onError` instead of throwing.
- Add tests covering a normal non-terminal server error and a malformed `error` message with no `error_event` payload.
- Add a changeset (`@elevenlabs/client` patch).

## Test plan

- [ ] `pnpm --filter @elevenlabs/client test` (new tests in `BaseConversation.test.ts` → `describe("error events")`)
- [ ] `pnpm --filter @elevenlabs/client lint` / typecheck

### DIFF
--- a/.changeset/guard-error-event-payload.md
+++ b/.changeset/guard-error-event-payload.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/client": patch
+---
+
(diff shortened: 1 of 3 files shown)


PaulAsjes (approved):
Thanks!

PR #900, opened 2026-07-24
Fix agent messages being deleted mid conversation

giannagerton:
Linear issue: [AWF-167: Widget messages replaced by new text during tool calls](https://linear.app/elevenlabs/issue/AWF-167/widget-messages-replaced-by-new-text-during-tool-calls)

Agent messages could be deleted and replaced mid-conversation. It happened when two replies in the same turn (one before a tool call and one after) shared an event_id. `buildDisplayTranscript` treated them as the same message and collapsed them into one bubble, so the second reply clobbered the first.

**Root cause**
The merge was gated by a `toolBetween` flag that only kept two same event_id replies apart if a tool request/response event showed up between them. But a server-side tool can run without emitting any client tool events, so the two replies end up back to back with nothing between them so `toolBetween` stays false and the second overwrites the first. (the `toolBetween` flag was introduced in this [PR](https://github.com/elevenlabs/packages/pull/861))

**Fix**
- Removed the `toolBetween` check entirely.
- Now we only fold an empty placeholder into the message that follows it. Two non-empty replies that share an event_id are always kept as separate bubbles.

This fix is safe because of how ingestion works: a streamed reply only ever lives in one transcript entry. start adds one empty row, each delta appends into that same row, and the final agent_response overwrites it in place (see onAgentChatResponsePart / onMessage in conversation.tsx). A partial and its final can never split into two rows. So two non-empty entries with the same event_id really are two different messages, and merging them would throw one away. Because the fix relies on that ingestion behavior, I added a browser test that verifies messages are consolidated into a single entry.

Before
gif_pr

After (There is a duplicate message issue now which I will fix in a separate PR but want to get this reviewed on it's own)
fix

---

> [!NOTE]
> **Medium Risk**
> Changes transcript bubble merging behavior for all conversations; wrong logic could duplicate bubbles or break streaming placeholders, but coverage targets the main regressions.
> 
> **Overview**
> Fixes **agent messages disappearing mid-conversation** when a turn sends multiple non-empty replies under the same `event_id` (common around tool calls, including server-side tools that never emit client tool events).
> 
> **`buildDisplayTranscript`** no longer collapses same-`eventId` agent bubbles using a `toolBetween` flag. It **only folds an empty placeholder** into the next same-turn message; **two non-empty entries stay separate bubbles**.
> 
> Unit tests cover adjacent pre/post-tool replies without tool transcript entries, and updated grouping expectations. A **`stream_consolidation`** MSW agent plus a browser test assert streamed partial + final still render as **one** bubble when ingestion overwrites in place.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 08ad14e134c2acfd9a7a9f150e7af4c0449ad0ec. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/convai-widget-core/src/index.test.ts
+++ b/packages/convai-widget-core/src/index.test.ts
@@ -143,6 +143,29 @@ describe("elevenlabs-convai", () => {
       .toBeInTheDocument();
   });
 
+  it("consolidates a streamed partial and its final into a single bubble", async () => {
+    // The reducer must consolidate a streamed message into one entry. The mock
+    // streams "partial" then delivers "full" as an agent_response for the same
+    // event_id, so "full" overwrites the partial in place.
+    setupWebComponent({
+      "agent-id": "stream_consolidation",
+      variant: "compact",
+    });
+
+    const textInput = page.getByRole("textbox", {
+      name: "Text message input",
+    });
+    await textInput.fill("go");
+    await userEvent.keyboard("{Enter}");
+
+    const finalMessage = page.getByText("full", { exact: true });
+    await expect.element(finalMessage).toBeInTheDocument();
+    expect(finalMessage.elements()).toHaveLength(1);
+    await expect
+      .element(page.getByText("partial", { exact: true }))
+      .not.toBeInTheDocument();
+  });
+
   it.each(Variants)(
     "$0 expandable variant should go through a happy path (text-only)",
     async variant => {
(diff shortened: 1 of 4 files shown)


kamilk-11 (approved):
LGTM

PR #921, opened 2026-08-07
feat(client): add self-hosted orchestrator session support

a1anfan:
Adds self-hosted orchestrator session support to `@elevenlabs/client`, matching the Python SDK's `OnPremInitiationData` (named `orchestratorConfig` here to avoid the `on*` callback-handler naming convention). Passing `orchestrator` to `Conversation.startSession` (or `createConnection`) routes the conversation WebSocket to a self-hosted orchestrator. The orchestrator is stateless, so the client sends an `enclave_setup_config` message (agent config, tools, optional Bedrock inference profile and post-call webhooks) before the standard `conversation_initiation_client_data` message; after that the protocol matches the cloud path. Orchestrator sessions are websocket-only.

Two connection details differ from the cloud path: the configured URL is used verbatim, and no WebSocket subprotocols are requested, because the orchestrator accepts the socket without subprotocol negotiation and browsers fail a connection whose requested subprotocol is not echoed back.

```ts
const conversation = await Conversation.startSession({
  orchestrator: {
    url: "wss://your-host/sagemaker/convai/conversation",
    agentConfig: config.agent_config,
    tools: config.tools_config,
  },
});
```

The React SDK picks this up through its session config passthrough with no changes; the widget and React Native can follow separately.

## Testing

- unit tests
- stub orchestrator e2e (Node WebSocket)
- browser e2e and manual testing against a deployed orchestrator

---

> [!NOTE]
> **Medium Risk**
> New connection path and session bootstrap protocol for private deployments; misconfiguration could fail sessions or expose webhook HMAC secrets in client-visible config (documented as test-only).
> 
> **Overview**
> Adds **experimental** self-hosted orchestrator sessions to `@elevenlabs/client` via an `orchestrator` option on `Conversation.startSession` / `createConnection`, aligned with the Python SDK’s on-prem initiation flow.
> 
> Sessions use **websocket only**: the client opens the configured URL as-is (no `convai` / bearer subprotocols) and, on connect, sends an **`enclave_setup_config`** message (agent config, overrides, tools, optional Bedrock profile and post-call webhooks) **before** the usual `conversation_initiation_client_data`. `createConnection` rejects mixing `orchestrator` with WebRTC, `signedUrl`, `conversationToken`, or `authorization`.
> 
> **`uploadFile`** is blocked for orchestrator sessions so file uploads are not sent to the ElevenLabs cloud API.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit c5c55d21b939bb688af7c9cc9d4df86ef0902c88. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/utils/BaseConnection.ts
+++ b/packages/client/src/utils/BaseConnection.ts
@@ -74,7 +74,13 @@ export type ConnectionType = "websocket" | "webrtc";
  */
 export type PostCallWebhookConfig = {
   url: string;
-  /** Secret the orchestrator uses to sign deliveries. The orchestrator rejects secrets shorter than 16 characters. */
+  /**
+   * Secret the orchestrator uses to sign deliveries. The orchestrator
+   * rejects secrets shorter than 16 characters. Anything set here is visible
+   * to the end user running the page, like the rest of the orchestrator
+   * session config; intended for testing against trusted deployments, not
+   * production use.
+   */
   hmacSecret?: string;
 };
 
--- a/packages/client/src/utils/BaseConnection.ts
+++ b/packages/client/src/utils/BaseConnection.ts
@@ -72,7 +72,7 @@ export type ConnectionType = "websocket" | "webrtc";
 /**
  * @experimental
  */
-export type OnPremWebhookConfig = {
+export type PostCallWebhookConfig = {
   url: string;
   /** Secret the orchestrator uses to sign deliveries. The orchestrator rejects secrets shorter than 16 characters. */
   hmacSecret?: string;
@@ -81,9 +81,9 @@ export type OnPremWebhookConfig = {
 /**
  * @experimental
  */
-export type OnPremConfig = {
+export type OrchestratorConfig = {
   /** WebSocket URL of the orchestrator, e.g. "wss://<host>/sagemaker/convai/conversation". */
-  conversationUrl: string;
+  url: string;
   /** Full agent definition, as exported from the ElevenLabs platform. */
   agentConfig?: Record<string, unknown>;
   /** Partial configs applied as overrides on top of the base agent config. */
@@ -93,9 +93,9 @@ export type OnPremConfig = {
   /** Strings appended to the system prompt at runtime. */
   promptKnowledgeBase?: string[];
   /** Called with the conversation transcript once the session ends. */
-  postCallTranscriptionWebhook?: OnPremWebhookConfig;
+  postCallTranscriptionWebhook?: PostCallWebhookConfig;
   /** Called with the conversation audio once the session ends. */
-  postCallAudioWebhook?: OnPremWebhookConfig;
+  postCallAudioWebhook?: PostCallWebhookConfig;
 };
 
 export type PublicSessionConfig = BaseSessionConfig & {
@@ -103,7 +103,7 @@ export type PublicSessionConfig = BaseSessionConfig & {
   connectionType?: ConnectionType;
   signedUrl?: never;
   conversationToken?: never;
-  onPremConfig?: never;
+  orchestratorConfig?: never;
 };
 
 export type PrivateWebSocketSessionConfig = BaseSessionConfig & {
@@ -111,7 +111,7 @@ export type PrivateWebSocketSessionConfig = BaseSessionConfig & {
   connectionType?: "websocket";
   agentId?: never;
   conversationToken?: never;
-  onPremConfig?: never;
+  orchestratorConfig?: never;
 };
 
 export type PrivateWebRTCSessionConfig = BaseSessionConfig & {
@@ -119,18 +119,18 @@ export type PrivateWebRTCSessionConfig = BaseSessionConfig & {
   connectionType?: "webrtc";
   agentId?: never;
   signedUrl?: never;
-  onPremConfig?: never;
+  orchestratorConfig?: never;
 };
 
 /**
  * @experimental
  */
-export type OnPremSessionConfig = BaseSessionConfig & {
+export type OrchestratorSessionConfig = BaseSessionConfig & {
   /**
    * Routes the session to a self-hosted orchestrator instead of the ElevenLabs cloud.
    * @experimental
    */
-  onPremConfig: OnPremConfig;
+  orchestratorConfig: OrchestratorConfig;
   connectionType?: "websocket";
   agentId?: never;
   signedUrl?: never;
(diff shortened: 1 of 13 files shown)


a1anfan:
cursor review

a1anfan:
cursor review

kraenhansen:
I would like to have a conversation about your thoughts around this API. Specially to see if any of these on-prem specifics could be easily expressed in the existing configuration options or if we should instead take these "top-level" and discriminate based on a separate boolean or string-literal field. I also don't particularly like that the `onPremConfig` starts with `on*` as this is a pattern we use for callback handlers already.

In any case these should be marked with the `@experimental` tsdoc tag to allow us changing these without the need for a major version bump.

packages/client/src/utils/BaseConnection.ts
```diff
@@ -69,32 +69,71 @@ export type BaseSessionConfig = {
 
 export type ConnectionType = "websocket" | "webrtc";
 
+export type OnPremWebhookConfig = {
+  url: string;
+  /** Secret the orchestrator uses to sign deliveries. The orchestrator rejects secrets shorter than 16 characters. */
+  hmacSecret?: string;
+};
+
+export type OnPremConfig = {
```
lagercat: For us to figure out how to generate these automatically with OpenAPI so we don't have  to keep updating it with new parameters.

a1anfan:
cursor review

a1anfan:
@kraenhansen, added @experimental tags (513365a), and went ahead and renamed "onPremConfig" to "orchestratorConfig" (05417b5) so you can see it concretely — the docs define the orchestrator as the gateway exposing the conversation WebSocket, and the wire types here already use ClientToOrchestratorEvent, so it felt like the natural term. also renamed the webhook type to PostCallWebhookConfig to match the Python SDK. happy to change it if you have a different direction in mind

re folding the fields into existing options: these fields change the connection itself rather than the conversation config since the URL is used verbatim, no convai subprotocol can be requested (i.e. browsers fail the connection if the server doesn't echo it, and self-hosted orchestrators don't negotiate), and an extra `enclave_setup_config` message precedes the normal initiation. the grouped object keeps those co-dependent fields together and mirrors the Python SDK's `OnPremInitiationData`. however lmk you feel strongly about a top-level discriminator shape!

kraenhansen:
I'd like some more time to dig into this in detail to see how this could potentially reuse part of the or replace the existing "top-level" config. Wondering if this should be encoded as a new `connectionType`?

kraenhansen:
Small naming note: since everything in `OrchestratorConfig` is inherently config, the `Config` suffix on the field itself (`orchestratorConfig`) reads as redundant — the type name already carries that. We drop the suffix on the field elsewhere, e.g. `connectionDelay: DelayConfig` (`BaseConnection.ts:65`), so `orchestrator: OrchestratorConfig` would be more consistent with that pattern.

kraenhansen:
Could you share more about the expected shape of entries in `toolsConfigList`? It's currently typed as `Record<string, unknown>[]` with no reference type, and the only shape hint is the test fixture (`{ type: "webhook" }`) — is this meant to match the platform's tool-export format exactly, and is there a schema we could reference or link to in the doc comment?

Same naming note as `orchestratorConfig`: `Config` and `List` both read as type-related rather than name-related (the `[]` and object shape already say "list of configs"), so `tools: Record<string, unknown>[]` would drop the redundancy — same logic would apply to `overrideAgentConfigList`.

kraenhansen:
One more: what happens if a caller sets both `overrides.agent.prompt` (replaces the prompt) and `orchestratorConfig.promptKnowledgeBase` (appends to it)? Since these act on the same underlying prompt via two different mechanisms/messages (`conversation_initiation_client_data` vs `enclave_setup_config`), it'd be good to spell out the interaction in the doc comments.

We have a precedent for a similar conflict with `textOnly` vs `overrides.conversation.textOnly` (`BaseConversation.ts:97-113`), though note that one only warns and picks a winner rather than failing — worth deciding whether `prompt`/`promptKnowledgeBase` should follow that same soft pattern, or fail fast client-side, since "replace" vs "append" feels like a more genuine logical conflict than two conflicting booleans.

a1anfan:
cursor review

a1anfan:
thanks for the thorough review @kraenhansen! all points addressed:

- **naming**: field is now `orchestrator`, plus `tools` and `agentConfigOverrides` ([64b1adc](https://github.com/elevenlabs/packages/commit/64b1adc8329172470abdc9d9c74f8c6331e4b024)). open to a better name for the last one.
- **`tools` shape**: the platform's tool export format, same as the Python SDK's `tools_config_list`. the orchestrator validates entries server-side and rejects the session on mismatch. now in the doc comment; no public schema yet.
- **`promptKnowledgeBase` vs `overrides.agent.prompt`**: checked the orchestrator source, they compose rather than conflict: one replaces the knowledge-base section, the other sets the prompt text. documented ([64b1adc](https://github.com/elevenlabs/packages/commit/64b1adc8329172470abdc9d9c74f8c6331e4b024)), so no warn/fail needed.
- **`hmacSecret`**: visible to the end user by design, like the rest of the session config. it signs deliveries to the customer's own webhook receiver, and whoever can set the secret can also set the URL, so nothing privileged leaks. doc note added ([c5c55d2](https://github.com/elevenlabs/packages/commit/c5c55d21b939bb688af7c9cc9d4df86ef0902c88)).
- **client SDK vs server SDK**: the browser is the real target: a page where a customer's evaluators call an agent inside their own VPC, with no ElevenLabs cloud in the path. Python parity is only because that was the sole reference implementation.
- **client_secrets**: agreed that's the right end state. the orchestrator is WebSocket-only today, no HTTP surface or storage layer to mint against, which is exactly why everything is `@experimental`, so we can migrate without a major bump.
- **connectionType** (your earlier question): kept it as a session option rather than a new `connectionType` because orchestrator sessions still use the websocket transport; what changes is the endpoint and the setup handshake. presence-of-exactly-one of `agentId` / `signedUrl` / `conversationToken` / `orchestrator` also matches the existing union.

also fixed two Bugbot findings on #922: file upload UI is hard-off for orchestrator sessions so file bytes can't leave the customer network ([3c47153](https://github.com/elevenlabs/packages/commit/3c47153a1fbe005e488636590e9c4b2cdf00330a)), and a `language` attribute honored against a declared supported set now behaves like the picker ([fe7e674](https://github.com/elevenlabs/packages/commit/fe7e6745ee02adb06c97210d40eae80a49dadeb2)). both verified on the wire against a stub orchestrator.

kraenhansen (approved):

PR #922, opened 2026-08-07
feat(widget): add self-hosted orchestrator support via orchestrator-url attribute

a1anfan:
Stacked on #921. Adds self-hosted orchestrator support to the Convai widget: setting the new `orchestrator-url` attribute connects the widget to a self-hosted orchestrator instead of the ElevenLabs cloud, and the optional `orchestrator-agent-config` attribute carries the exported agent configuration JSON (both the `agent_config_dict`/`tools_config_list` and `agent_config`/`tools_config` key spellings are accepted, plus an optional top-level `bedrock_inference_profile`). The parsed config feeds the client SDK's `orchestrator` session config from #921; the connection is forced to websocket, since self-hosted orchestrators only expose the conversation WebSocket.

Because self-hosted deployments have no HTTP API to serve a widget appearance config, orchestrator sessions skip the config fetch entirely and use a built-in default appearance (transcript and text input enabled), which the existing attributes and `override-config` can still override. The file upload button stays hidden by default, so no request leaves the customer network. Both changes are inert unless `orchestrator-url` is set; cloud behavior is untouched.

## Testing

- unit tests
- build, lint, check-types across widget-core and widget-embed
- browser e2e and manual testing against a deployed orchestrator

---

> [!NOTE]
> **Medium Risk**
> Changes core session and config wiring and how connections are established; mistakes could break cloud sessions or mis-route self-hosted traffic, though cloud paths stay gated behind `orchestrator-url`.
> 
> **Overview**
> Adds **experimental self-hosted orchestrator** support to the Convai widget via new `orchestrator-url` and optional `orchestrator-agent-config` attributes (minor bump in changeset). When `orchestrator-url` is set, it **takes precedence** over `agent-id` / `signed-url`, skips the ElevenLabs widget config HTTP fetch, and builds a **websocket-only** client `orchestrator` session from parsed export JSON (`parseOrchestratorConfig`, with unit tests).
> 
> Orchestrator mode uses a **built-in default appearance** (still overridable via existing attributes / `override-config`), exposes **`languageOverride`** for session agent language when overrides apply, and **turns off cloud-only UI** (file upload and end-of-call feedback). `OrchestratorProvider` is wired into the main widget and dev playground trees.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit afcc6d2dca353e3c3c15bef0461f9f0d94b913c7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


a1anfan:
cursor review

a1anfan:
cursor review

a1anfan:
cursor review

a1anfan:
cursor review

a1anfan:
cursor review

packages/convai-widget-core/src/contexts/session-config.tsx
```diff
@@ -80,28 +102,66 @@ export function SessionConfigProvider({
...
       },
     } as const satisfies Partial<SessionConfig | AudioWorkletConfig>;
 
+    if (orchestratorUrl.value) {
+      const orchestrator = parsedOrchestratorConfig.value;
+      if (!orchestrator) {
+        return null;
+      }
+      // Boilerplate language="en" with no supported set must not override the agent config's own language.
+      const resolvedLanguage = language.value.languageCode;
```
kalvin807: I think this can burn down to just this logic

```
const languageOverride = isValidLanguage(overrideLanguage.value)
  ? overrideLanguage.value
  : undefined;
```

kalvin807 (commented):
logic make sense. not a fan of this structure (session-config is already too long and complex). can we isolate orchestrator logic into a context? 

```ts
import { computed, ReadonlySignal, useSignalEffect } from "@preact/signals";
import { ComponentChildren } from "preact";
import { createContext, useMemo } from "preact/compat";
import type { OrchestratorConfig } from "@elevenlabs/client";

import { useAttribute } from "./attributes";
import { useContextSafely } from "../utils/useContextSafely";
import { parseOrchestratorConfig } from "../utils/parseOrchestratorConfig";

const OrchestratorContext =
  createContext<ReadonlySignal<OrchestratorConfig | null> | null>(null);

export function OrchestratorProvider({ children }: { children: ComponentChildren }) {
  const url = useAttribute("orchestrator-url");
  const agentConfig = useAttribute("orchestrator-agent-config");
  const agentId = useAttribute("agent-id");
  const signedUrl = useAttribute("signed-url");

  const value = useMemo(
    () =>
      computed(() =>
        url.value ? parseOrchestratorConfig(url.value, agentConfig.value) : null
      ),
    []
  );

  useSignalEffect(() => {
    if (url.value && (agentId.value || signedUrl.value)) {
      console.warn(
        "[ConversationalAI] orchestrator-url takes precedence; agent-id and signed-url are ignored"
      );
    }
  });

  return (
    <OrchestratorContext.Provider value={value}>{children}</OrchestratorContext.Provider>
  );
}

export function useOrchestrator() {
  return useContextSafely(OrchestratorContext);
}

export function useIsOrchestratorSession() {
  const orchestrator = useOrchestrator();
  return useComputed(() => orchestrator.value !== null);
}
```

kalvin807:
also tagging @giannagerton for widget related changes

a1anfan:
cursor review

a1anfan:
both done, thanks for the push - this ended up much cleaner

orchestrator context (266a924): went with your shape - OrchestratorProvider / useOrchestrator in orchestrator-config.tsx, mounted above WidgetConfigProvider, and the widget-config upload/feedback gates consume it too. one deviation: the context exposes enabled (attribute set) separately from the parsed config, because deriving "orchestrator session" from config-non-null would let an orchestrator url with an invalid agent config fall through to a cloud session when agent-id is also set.

language override (732f553, e1345ec): your one-liner drops the picker and browser matching (`language.value` already folds in override-language → picker → browser), and cursor caught two divergence bugs in my first simplification, both from the same root: for orchestrator sessions the widget appearance default isn't the agent's real language. final state is simpler than both: language-config exposes one `languageOverride` signal - "send the resolved language iff override-language is set or a supported set is declared", i.e. whenever the picker is visible the conversation follows it. no `languageChosen` flag, survives reloads, and bare boilerplate language="en" still never overrides the agent config. session-config is down to language: `languageOverride.value`.

one behavior note: with a declared supported set the override is now also sent on first load (previously silent until something diverged). this is deliberate because the picker was already showing a language the agent wasn't speaking

packages/convai-widget-core/src/contexts/language-config.tsx
```diff
@@ -115,25 +116,31 @@ export function LanguageConfigProvider({
...
-  const value = useMemo(
-    () => ({
-      language: computed(() =>
-        isValidLanguage(overrideLanguageAttribute.value)
-          ? Languages[overrideLanguageAttribute.value]
-          : isValidLanguage(languageCode.value) &&
-              supportedOverrides.value.includes(languageCode.value)
-            ? Languages[languageCode.value]
-            : Languages[widgetConfig.value.language]
+  const value = useMemo(() => {
```
kalvin807, line 119: nit: add comment to explain this nested ternary.

kalvin807 (approved):
lgtm

kraenhansen (approved):

packages/convai-widget-core/src/contexts/language-config.tsx
a1anfan: [added](https://github.com/elevenlabs/packages/pull/922/changes/73f7a0a77babd9c6fdb1307b2f1e003d5d1dbc6a)

PR #934, opened 2026-08-12
Add webRtc.iceTransportPolicy session option for WebRTC connections

a1anfan:
Adds an optional `webRtc` object to the session config with a single primitive field, `iceTransportPolicy` (`"all" | "relay"`). Setting `"relay"` restricts ICE to TURN relay candidates, for networks that drop direct UDP flows mid-call. Until now the only way to control this was monkey-patching the global `RTCPeerConnection` before session start. When unset, connect behavior is unchanged.

The public API surface stays independent of livekit-client and free of DOM types; the mapping to livekit's `room.connect()` `rtcConfig` lives inside `WebRTCConnection`. The option flows to `@elevenlabs/react` and `@elevenlabs/react-native` automatically through the shared config types. livekit-client applies the policy to both the publisher and subscriber peer connections and keeps server-provided ICE servers.

## Testing

- unit tests
- build + lint across client, react, and react-native

### DIFF
--- a/.changeset/rtc-config-passthrough.md
+++ b/.changeset/rtc-config-passthrough.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/client": minor
+---
+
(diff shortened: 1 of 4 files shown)


a1anfan:
cursor review

kraenhansen (changes requested):
Reached out over Slack

kraenhansen (approved):

PR #912, opened 2026-08-03
Add onRichContent callback

giannagerton:
## Summary
Adds an experimental `onRichContent` callback so the SDK can hand off agent-driven UI components (e.g. an item card) to whoever's consuming it.

- New `RichContentClientEvent`/`RichContentData` schema in `@elevenlabs/types` (AsyncAPI + generated types)
- `onRichContent` callback in `@elevenlabs/client`, dispatched from `BaseConversation` the same way every other server-pushed event is — no reply sent back, the agent's turn doesn't wait on it
- Re-exported through `@elevenlabs/react`'s `HookCallbacks`

Backs the embedded widget, which is the only client the server sends components to today — this callback won't fire for other consumers yet. `props` is agent-authored; treat it as untrusted when rendering.

## Summary

[RFC](https://docs.google.com/document/d/1ZMxvMifKVaUxnDSLHVTI3JWxrfmC0OQzguC-d_Uf32k/edit?usp=sharing)

Adds an experimental `onRichContent` callback so the SDK can hand off agent driven UI to the widget.

- New `RichContentClientEvent`/`RichContentData` schema in `@elevenlabs/types` 
- `onRichContent` callback in `@elevenlabs/client`, dispatched from `BaseConversation`. No reply sent back, the agent's turn doesn't wait on it
- Re-exported through `@elevenlabs/react`'s `HookCallbacks`

---

> [!NOTE]
> **Medium Risk**
> New experimental surface that forwards untrusted `props` to app code; misuse when rendering could introduce XSS, but core conversation and auth paths are unchanged.
> 
> **Overview**
> Adds **`rich_content`** to the agent WebSocket contract and wires it through **`@elevenlabs/types`**, **`@elevenlabs/client`**, and **`@elevenlabs/react`** as an experimental **`onRichContent`** callback.
> 
> When the server sends a display component (e.g. **`item_card`**), the client invokes the callback with **`{ rich_content_id, component, props, event_id }`**. Unlike **`client_tool_call`**, the SDK does **not** send a response, so the agent turn is not blocked on the client.
> 
> **`BaseConversation`** dispatches **`rich_content`** in **`onMessage`** via **`handleRichContent`**; tests cover payload forwarding, no **`sendMessage`**, and missing callback. The changeset notes **`props`** should be treated as agent-authored and untrusted when rendering into a document.
> 
[...]

PR #935, opened 2026-08-12
[Widget] Render agent rich content

giannagerton:
Adds the widget-side consumer of the `onRichContent` SDK callback (#912). Adds rich content rendering to the chat widget. The agent can display components inline in the transcript starting with quick reply buttons. It's inert on merge, since nothing on the server emits the event yet.

Starting with buttons rather than cards, for two reasons:
- ButtonGroup is the primitive other components build on. A card or carousel renders its own buttons through the same component.
- Going through the customer analysis, clickable links and guided options came up far more often than product cards.

Screenshot 2026-08-12 at 2 39 13 PM

## Button types

| Type | Behaviour |
|---|---|
| `message` | Sends its text as the user's own turn |
| `link` | Opens its url in a new tab |

Max 3 buttons per row. Buttons that can run actions directly to follow.

---

> [!NOTE]
> **Medium Risk**
> New user-facing transcript paths and outbound links/message sends, mitigated by strict https-only link validation and safe fallbacks for bad payloads.
> 
> **Overview**
> Adds **inline rich content** in the convai widget transcript, wired from the SDK **`onRichContent`** callback into new **`rich_content`** transcript entries and display rows.
> 
> The first supported component is **`buttons`**: **message** buttons call **`sendUserMessage`** with the payload text; **link** buttons use an extended **`Button`** (`as="a"`, **`outline`**, `target="_blank"`). Props are validated with **`zod/mini`** (caps, trimming, **https-only** links); unknown or invalid payloads show **`rich_content_unavailable`** instead of breaking the UI.
> 
> Display transcript behavior is covered by new tests so rich content stays in **arrival order** relative to agent replies, tool-status placeholders, and the typing indicator.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 73cdfb92f79f3344ad538df3527bf22019ba53fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/convai-widget-core/src/rich-content/ButtonGroup.tsx
+++ b/packages/convai-widget-core/src/rich-content/ButtonGroup.tsx
@@ -1,0 +1,58 @@
+import {
+  Button,
+  buttonClassName,
+  buttonLabelClassName,
+} from "../components/Button";
+import { useConversation } from "../contexts/conversation";
+
+export interface MessageButton {
+  type: "message";
+  label: string;
+  message: string;
+}
+
+export interface LinkButton {
+  type: "link";
+  label: string;
+  link: string;
+}
+
+export type RichContentButton = MessageButton | LinkButton;
+
+export interface ButtonGroupProps {
+  buttons: RichContentButton[];
+}
+
+export function ButtonGroup({ buttons }: ButtonGroupProps) {
+  const { sendUserMessage, status } = useConversation();
+  const disabled = status.value !== "connected";
+
+  if (buttons.length === 0) return null;
+
+  return (
+    <div className="flex flex-wrap gap-2">
+      {buttons.map((button, index) =>
+        button.type === "link" ? (
+          <a
+            key={index}
+            href={button.link}
+            target="_blank"
+            rel="noopener noreferrer"
+            className={buttonClassName("outline")}
+          >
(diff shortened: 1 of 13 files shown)


packages/convai-widget-core/src/rich-content/ButtonGroup.tsx
```diff
@@ -0,0 +1,58 @@
...
+  const { sendUserMessage, status } = useConversation();
+  const disabled = status.value !== "connected";
+
+  if (buttons.length === 0) return null;
+
+  return (
+    <div className="flex flex-wrap gap-2">
+      {buttons.map((button, index) =>
+        button.type === "link" ? (
+          <a
```
kamilk-11: Maybe we could add support for `as="a"` directly to the `Button`? Seems like there are gonna be more places where this is desired

packages/convai-widget-core/src/rich-content/validate.ts
kamilk-11, line 1: Maybe we should bring in something like [Zod Mini](https://zod.dev/packages/mini)?
I think the size footprint should be negligible

kamilk-11 (approved):
Nice!

PR #936, opened 2026-08-13
feat(connection): add handlers to monitor incoming/outgoing messages

mat-sz:
Linear issue: [AWF-256](https://linear.app/elevenlabs/issue/AWF-256/add-callbacks-for-incomingoutcoming-messages-to-elevenlabsclient)

Added `onIncomingEvent` and `onOutgoingEvent` handlers that will be called every time a message is being received or being sent, regardless of the message type. This will make monitoring the payloads easier.

---

> [!NOTE]
> **Low Risk**
> Additive observability callbacks on the message path; existing dispatch and send behavior are unchanged aside from optional listener invocation.
> 
> **Overview**
> Adds **`onIncomingEvent`** and **`onOutgoingEvent`** conversation callbacks so apps can observe every socket payload without relying on type-specific handlers or `onDebug`.
> 
> **Incoming:** `BaseConversation` invokes `onIncomingEvent` with the raw parsed event before the existing `switch` dispatch, so it runs for all server messages (including ones that only hit `onDebug` today).
> 
> **Outgoing:** `BaseConnection` gains `onOutgoingMessage` with the same early-registration queue flush pattern as incoming messages. `WebSocketConnection` and `WebRTCConnection` call `handleOutgoingMessage` when `sendMessage` runs (WebRTC still skips `user_audio_chunk` on the wire, unchanged). `BaseConversation` forwards those to `onOutgoingEvent`.
> 
> The React hook surface, `CALLBACK_KEYS`, and the agent testbench event spy list include the new callbacks. Unit tests cover incoming forwarding and outgoing reporting on both transports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 1b85e4a62ab46aaa443452016541f396a98cb52b. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/examples/agent-testbench/src/components/agent-page.tsx
+++ b/examples/agent-testbench/src/components/agent-page.tsx
@@ -40,6 +40,8 @@ const EVENT_METHOD_NAMES = [
   "onGuardrailTriggered",
   "onPing",
   "onDebug",
+  "onIncomingEvent",
+  "onOutgoingEvent",
 ] satisfies (keyof PartialOptions)[];
 
 export function AgentPage({ agent }: AgentPageProps) {
--- a/packages/client/src/BaseConversation.test.ts
+++ b/packages/client/src/BaseConversation.test.ts
@@ -10,6 +10,7 @@ import type { BaseConnection } from "./utils/BaseConnection.js";
 const noopConnection = {
   conversationId: "test-conversation-id",
   onMessage: () => {},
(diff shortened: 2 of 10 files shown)


kamilk-11 (approved):
LGTM

PR #937, opened 2026-08-13
[Widget] Fix language dropdown position

giannagerton:
The language dropdown rendered detached from its trigger when the widget was embedded inside a CSS container-query element. Floating UI treats a container-type ancestor as a containing block for fixed descendants while browsers don't. Adding `transform: translate(0)` to `:host` gives the widget its own containing block making the position independent of the embedding page.

Before:
Screenshot 2026-08-13 at 11 19 08 AM

After:
Screenshot 2026-08-13 at 11 19 30 AM

### DIFF
--- a/.changeset/lucky-moons-anchor.md
+++ b/.changeset/lucky-moons-anchor.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/convai-widget-core": patch
+---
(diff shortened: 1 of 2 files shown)


kamilk-11 (approved):
Thanks!

PR #939, opened 2026-08-15
fix(client): fail WebRTC setup when the initiation payload cannot be published

### LINKED ISSUE #916: [@elevenlabs/client][WebRTC] conversation-init publish failures are swallowed, allowing silent sessions
## Summary

`WebRTCConnection.create()` sends the required
`conversation_initiation_client_data` through the generic best-effort
`sendMessage()` path.

That helper:

1. returns successfully when the room is no longer connected, and
2. catches and swallows `localParticipant.publishData()` rejection.

As a result, `create()` can resolve and `Conversation.startSession()` can fire
`onConnect` even though the server never received the initialization payload.
The LiveKit room and microphone can remain active while no usable agent
conversation is initialized.

The same code path is present on current `main` (`@elevenlabs/client` reports
version 1.17.0):
https://github.com/elevenlabs/packages/blob/main/packages/client/src/utils/WebRTCConnection.ts

## Reproduction

A deterministic unit test can mock the LiveKit `Room` so connection and
microphone setup succeed, then exercise either case immediately before the init
send:

- mark the connection disconnected, or
- make `room.localParticipant.publishData()` reject.

### Actual

`WebRTCConnection.create()` resolves because `sendMessage()` returns or catches.
A higher layer can report the conversation as connected even though no
conversation metadata or agent activity arrives.

### Expected

The mandatory initialization publish should reject setup. The existing
`create()` catch already disconnects the room and rethrows, so the caller's
normal connection-error path can recover.

Normal mid-session messages can remain best effort. Making `sendMessage()`
globally throw would be risky because other SDK callers use it fire-and-forget.
The strict behavior only needs to apply to this initialization publish.

## Observed impact

We observed production sessions with user audio but no conversation metadata,
transcript turns, or model-token usage, together with empty dynamic variables
and null overrides. This is consistent with the failure mode above, although
provider-side ingestion or reporting loss cannot be excluded from production
records alone.

No customer identifiers or conversation IDs are included here.

## Versions

- observed with `@elevenlabs/client@1.8.1` through
  `@elevenlabs/react-native@1.2.3`
- verified that current repository `main` / client version 1.17.0 still uses
  the same best-effort init-send path

chinmayv095:
Fixes #916.

## Problem

`WebRTCConnection.create()` sends the mandatory `conversation_initiation_client_data` through `sendMessage()`, which is best effort in two ways:

- it returns early, with only a `console.warn`, when the room is no longer connected
- it catches and swallows a `localParticipant.publishData()` rejection

Either case lets `create()` resolve. `Conversation.startSession()` then fires `onConnect` and the caller reports a connected session, while the LiveKit room and the microphone stay live for a conversation the server never initialized — audible to the user as a session that connects and then never speaks.

## Fix

Send the initiation payload through a path that surfaces those failures instead of absorbing them. `create()`'s existing `catch` already calls `room.disconnect()` and rethrows, so the caller's normal connection-error handling recovers with no new error plumbing.

Ordinary mid-session sends are unchanged and stay best effort — that distinction is the point, so it has its own test.

## Tests

Three added to `WebRTCConnection.test.ts`, driving the mocked LiveKit `Room` so connect and mic setup succeed and only the initiation send fails:

| Test | Without this change |
|---|---|
| `fails setup when the initiation payload cannot be published` | fails — `create()` resolves |
| `fails setup when the room disconnects before the payload is sent` | fails — `create()` resolves |
| `keeps mid-session sends best effort when publishing fails` | passes — guards against over-fixing |

The first also asserts `room.disconnect()` ran, so a failed setup cannot leave the room and microphone live.

Full `@elevenlabs/client` suite: 182 passing (179 before). `lint:es`, `lint:prettier` and `tsc --build` clean. Changeset included per `agents.md`.

## Note

No conflict with #934, which is also in this file — it changes `room.connect()` around line 318, while this touches the initiation send and `sendMessage`.

---

> [!NOTE]
> **Medium Risk**
> Touches WebRTC session bootstrap and connection error semantics; behavior change is intentional but affects when callers see connected vs failed sessions.
> 
> **Overview**
> **WebRTC `create()` no longer reports success when mandatory `conversation_initiation_client_data` never reaches the server.** That payload is sent through `sendRequiredMessage()` instead of best-effort `sendMessage()`, so a disconnected room or a failed `publishData()` rejects setup and flows through the existing `catch` that calls `room.disconnect()` and rethrows.
> 
> Ordinary mid-session data messages still use `sendMessage()` (warn/return or log and swallow errors). Publishing is factored into shared `publishMessage()` used by both paths.
> 
> Tests cover publish failure during setup (including `disconnect`), disconnect before send without calling `publishData`, and unchanged best-effort behavior after a successful `create()`. Patch changeset for `@elevenlabs/client`.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f4eec6b2d562e04a3db421596968c60e63d647fc. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/webrtc-required-init-publish.md
+++ b/.changeset/webrtc-required-init-publish.md
@@ -1,0 +1,15 @@
+---
+"@elevenlabs/client": patch
+---
+
+Fail WebRTC session setup when the conversation initiation payload cannot be
+published, instead of resolving onto a session the server never initialized.
+
+`WebRTCConnection.create()` sent `conversation_initiation_client_data` through
+the same best-effort path as ordinary messages, which returns early when the
+room is no longer connected and swallows a `publishData()` rejection. Setup
+could therefore succeed, and `onConnect` fire, while the room and microphone
+stayed live for a conversation that was never initialized. The mandatory
+initiation send now surfaces those failures so the existing `create()` error
+path disconnects the room and reports the failure to the caller. Mid-session
+sends remain best effort.
--- a/packages/client/src/utils/WebRTCConnection.test.ts
+++ b/packages/client/src/utils/WebRTCConnection.test.ts
@@ -498,4 +498,109 @@ describe("WebRTCConnection", () => {
       setWebRTCAudioAdapterFactory(() => new WebAudioAdapter());
     }
   });
+
+  describe("conversation initiation payload", () => {
+    let mockRoom: any;
+
+    beforeEach(() => {
+      mockRoom = new Room() as any;
+
+      (mockRoom.once as ReturnType<typeof vi.fn>).mockImplementation(
+        (event: string, callback: () => void) => {
+          if (event === "signalConnected") {
+            queueMicrotask(callback);
+          }
+        }
+      );
+      (
+        mockRoom.localParticipant.getTrackPublication as ReturnType<
+          typeof vi.fn
+        >
+      ).mockReturnValue(undefined);
+    });
+
+    it("fails setup when the initiation payload cannot be published", async () => {
+      (mockRoom.on as ReturnType<typeof vi.fn>).mockImplementation(
+        (event: string, callback: () => void) => {
+          if (event === "connected") {
+            queueMicrotask(callback);
+          }
+        }
+      );
+      (
+        mockRoom.localParticipant.publishData as ReturnType<typeof vi.fn>
+      ).mockRejectedValueOnce(new Error("publish failed"));
+
+      await expect(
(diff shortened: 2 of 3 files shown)


kraenhansen (approved):

kraenhansen:
Thanks a lot @chinmayv095 for the fix 👍

PR #943, opened 2026-08-17
fix(client): pass workletPaths through on the WebRTC connection path

### LINKED ISSUE #649: WebRTC path ignores workletPaths config, preventing CSP-safe AudioWorklet loading
## Summary

The `workletPaths` option in `AudioWorkletConfig` is not passed through on the WebRTC connection path, making it impossible to self-host AudioWorklet files when using `conversationToken` (WebRTC mode). This forces apps with strict CSP to add `blob:` and `data:` to `script-src`, which weakens XSS protection.

## Versions

- `@elevenlabs/client`: 1.1.1
- `@elevenlabs/react`: 1.0.2

## Current Behavior

When using `conversationToken` (which routes through `ConnectionFactory` → `WebRTCConnection`), the `WebRTCConnection.js` calls:

```js
loadRawAudioProcessor(audioContext.audioWorklet)
// No path argument passed ↑
```

This means the `createWorkletModuleLoader` fallback always uses `blob:` / `data:` URIs, even when `workletPaths.rawAudioProcessor` was provided in the session options.

## Expected Behavior

`WebRTCConnection` should pass the `workletPaths` config through to the worklet loaders, just like the WebSocket path does via `VoiceSessionSetup` → `input.js` / `output.js`.

## Why This Matters

Apps with strict Content Security Policy headers cannot allow `blob:` and `data:` in `script-src` without weakening XSS protection. The SDK already has the right abstraction (`createWorkletModuleLoader` checks for a `path` argument first), but the WebRTC path doesn't use it.

## Workaround

Currently the only workaround is adding `blob: data:` to `script-src` in CSP headers, which is a security trade-off.

## Affected Files

- `dist/utils/WebRTCConnection.js` — calls `loadRawAudioProcessor` without path
- `dist/utils/createWorkletModuleLoader.js` — already supports path-first loading (line 9)
- `dist/utils/input.js` / `dist/utils/output.js` — correctly pass paths on the WebSocket path

chinmayv095:
Fixes #649

## The problem

`workletPaths` exists so apps under a strict CSP can self-host the AudioWorklet files instead of allowing `blob:` and `data:` in `script-src`. It works on the WebSocket path and is ignored on the WebRTC path.

`WebAudioAdapter.setupOutputAnalysis` — the WebRTC output-capture path that backs the `onAudio` callback — called the loader with no path:

```ts
await loadRawAudioProcessor(audioContext.audioWorklet);
```

while `MediaDeviceInput.create` on the WebSocket path passes one:

```ts
await loadRawAudioProcessor(context.audioWorklet, workletPaths?.rawAudioProcessor);
```

`createWorkletModuleLoader` already prefers a path when given one, so the abstraction was in place — the WebRTC path just never reached it. Concretely: a session started with `conversationToken` (or any `connectionType: "webrtc"`) plus `workletPaths.rawAudioProcessor` still constructs a `Blob`, calls `URL.createObjectURL`, and loads the worklet from a `blob:` URL. Under `script-src 'self'` that load is refused and output capture fails; the only escape is to weaken the CSP.

The config was reaching `createConnection` at runtime — `VoiceSessionSetup` passes the full `Options` object — it was simply never read on this path.

### One extra consequence worth flagging

`createWorkletModuleLoader` caches by worklet **name**, not by URL:

```ts
const cachedUrl = URLCache.get(name);
if (cachedUrl) return worklet.addModule(cachedUrl);
```

Both paths load the worklet named `rawAudioProcessor`. So a WebRTC session that fell back to a blob cached that blob URL under `rawAudioProcessor`, and a *later* WebSocket session in the same document — one that does pass `workletPaths` — got the cached `blob:` URL instead of its own path. The CSP-safe path was defeated for the rest of the page lifetime, not just for the WebRTC session. This PR removes the source of that poisoned entry; I've left the cache keying itself alone as a separate concern.

## The fix

Thread `workletPaths` from the connection config to the adapter:

- `WebRTCConnection` accepts `workletPaths` in its config, stores it, and passes it to `setupOutputAnalysis`. This matches how `onDebug` already reaches `WebRTCConnectionConfig` without being part of `SessionConfig`.
- `WebRTCAudioAdapter.setupOutputAnalysis` takes an optional fourth `workletPaths` argument.
- `WebAudioAdapter` forwards `workletPaths?.rawAudioProcessor` to `loadRawAudioProcessor`.

The new parameter is optional, so any existing adapter implementation still satisfies the interface. The React Native adapter needs no change — it does no worklet loading.

Output playback on WebRTC goes through `HTMLAudioElement` and input goes through LiveKit, so `rawAudioProcessor` in `setupOutputAnalysis` is the only worklet load on this path. `audioConcatProcessor` and `libsampleratePath` are WebSocket-only and untouched.

## Verification

Six new tests, four in `webAudioAdapter.test.ts` and two in `WebRTCConnection.test.ts`.

Both halves were checked fail-first by stashing only the source file:

- Stashing `webAudioAdapter.ts` → **3 failed / 1 passed**. The one that passes is the control asserting the capture graph is still wired the same way (`source → analyser → worklet`, `setFormat` posted, analyser returned) — it must hold in both states, and it does.
- Stashing `WebRTCConnection.ts` → **2 failed / 9 passed**, i.e. exactly the two new connection-level tests fail and every existing test in that file still passes.

The negative cases are deliberate: one asserts `undefined` is forwarded when no `workletPaths` are configured, and one asserts `undefined` when `workletPaths` carries only `audioConcatProcessor`, so the change can't silently pass the wrong processor's path.

Full `@elevenlabs/client` suite: **200 passed / 16 files**, up from 194 by exactly the six added. `turbo lint` and `check-types` pass across all 9 packages.

---

> [!NOTE]
> **Low Risk**
> Narrow, optional config threading on the WebRTC audio-capture path with backward-compatible adapter API and focused tests; no auth or data-model changes.
> 
> **Overview**
> **WebRTC sessions now honor `workletPaths` for remote output capture**, matching behavior that already existed on the WebSocket path. Apps with a strict CSP can self-host `rawAudioProcessor` instead of loading it from `blob:`/`data:` URLs when using `connectionType: "webrtc"`.
> 
> `WebRTCConnectionConfig` accepts optional `workletPaths`, stores them on the connection, and passes them into `WebRTCAudioAdapter.setupOutputAnalysis`. The web `WebAudioAdapter` forwards `workletPaths?.rawAudioProcessor` to `loadRawAudioProcessor` when wiring the output capture graph for the `onAudio` callback.
> 
> Six tests cover adapter forwarding (including unrelated processor keys), unchanged capture wiring, and connection-level propagation when an agent audio track is subscribed.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 60ee4da73e17829a36404c588b0ba6894abd8b01. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/webrtc-worklet-paths.md
+++ b/.changeset/webrtc-worklet-paths.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/client": patch
+---
+
+Pass `workletPaths` through on the WebRTC connection path, so self-hosted AudioWorklet files are used for output capture instead of falling back to `blob:`/`data:` URLs under a strict CSP.
--- a/packages/client/src/WebRTCAudioAdapter.ts
+++ b/packages/client/src/WebRTCAudioAdapter.ts
@@ -1,4 +1,5 @@
 import type { RemoteAudioTrack } from "livekit-client";
+import type { AudioWorkletConfig } from "./BaseConversation.js";
 import type { FormatConfig } from "./utils/BaseConnection.js";
 import type { VolumeProvider } from "./utils/volumeProvider.js";
 
@@ -39,11 +40,14 @@ export interface WebRTCAudioAdapter {
   /**
    * Set up output volume analysis and raw audio capture from a remote track.
    * @param onAudioData Called with captured audio data for the `onAudio` callback.
+   * @param workletPaths Optional self-hosted worklet paths, so implementations
+   * that load an AudioWorklet can avoid `blob:`/`data:` URLs under a strict CSP.
    */
   setupOutputAnalysis(
     track: RemoteAudioTrack,
     format: FormatConfig,
-    onAudioData: (audioData: ArrayBuffer, maxVolume: number) => void
+    onAudioData: (audioData: ArrayBuffer, maxVolume: number) => void,
+    workletPaths?: AudioWorkletConfig["workletPaths"]
   ): Promise<AnalysisResult>;
 
   /** Set playback volume (0-1) for all managed audio elements. */
--- a/packages/client/src/platform/web/webAudioAdapter.test.ts
+++ b/packages/client/src/platform/web/webAudioAdapter.test.ts
@@ -1,0 +1,137 @@
+import { afterEach, describe, expect, it, vi } from "vitest";
+
+vi.mock("./rawAudioProcessor.generated.js", () => ({
+  loadRawAudioProcessor: vi.fn(),
+}));
+
+import type { RemoteAudioTrack } from "livekit-client";
+import type { FormatConfig } from "../../utils/BaseConnection.js";
+import { loadRawAudioProcessor } from "./rawAudioProcessor.generated.js";
+import { WebAudioAdapter } from "./webAudioAdapter.js";
+
+const OUTPUT_FORMAT: FormatConfig = { format: "pcm", sampleRate: 48000 };
+
+function installAudioEnvironment() {
+  const analyser = {
+    fftSize: 0,
+    smoothingTimeConstant: 0,
+    frequencyBinCount: 1024,
+    connect: vi.fn(),
+    getByteFrequencyData: vi.fn(),
+  };
+  const source = { connect: vi.fn() };
+  const worklet = {
+    port: {
+      onmessage: null,
+      postMessage: vi.fn(),
+    },
+  };
+  const audioContext = {
+    audioWorklet: {},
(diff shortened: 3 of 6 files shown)


kraenhansen (approved):

kraenhansen:

PR #944, opened 2026-08-17
fix(client): tell React Native users to import @elevenlabs/react-native

### LINKED ISSUE #604: Throw if @elevenlabs/client is used in React Native without @elevenlabs/react-native setup
## Problem

After the latest SDK refactor (currently on the `next` branch), `@elevenlabs/client` is platform-agnostic but requires `@elevenlabs/react-native` to be imported first when running in a React Native environment. The `@elevenlabs/react-native` package performs essential side-effects:

- Polyfills WebRTC globals via `registerGlobals()`
- Configures native `AudioSession`
- Registers a React Native `VoiceSessionSetupStrategy` via `setSetupStrategy()`

If a user imports `@elevenlabs/client` (or `@elevenlabs/react`) directly in React Native without importing `@elevenlabs/react-native` first, they'll get confusing runtime errors (e.g. missing WebRTC globals, wrong audio session config).

## Proposal

Add a runtime check in `@elevenlabs/client` that detects a React Native environment (e.g. `navigator.product === 'ReactNative'` or `typeof global.HermesInternal !== 'undefined'`) and throws a clear error if the setup strategy hasn't been registered yet.

Something like:

```
Error: It looks like you're using @elevenlabs/client in React Native without importing @elevenlabs/react-native first. Please add `import '@elevenlabs/react-native'` before any other ElevenLabs imports.
```

## Where to check

The check should run at the point where a voice session is being started (e.g. in `VoiceConversation` or `ConnectionFactory`), not at import time.

chinmayv095:
Fixes #604

## The problem

`@elevenlabs/client` is platform-agnostic and depends on `@elevenlabs/react-native` being imported first in a React Native app — that package polyfills the WebRTC globals via `registerGlobals()`, configures the native `AudioSession`, and registers a React Native `VoiceSessionSetupStrategy`.

There is already a guard for a missing strategy in `VoiceConversation.startSession`, but its message only describes the browser fix:

```
No voice session setup strategy registered. Import the platform-specific entry point (e.g. @elevenlabs/client via the "browser" export).
```

`src/index.ts` registers no strategy — only `platform/web/index.ts` does, at import time. So a React Native app that resolves the package's `default` export rather than `browser` hits exactly this error, and the message tells it to do the one thing that will not help: importing the browser entry point in React Native pulls in the web setup path instead of the native one. The correct action is `import "@elevenlabs/react-native"`.

## The fix

As suggested in the issue, the check runs when a voice session starts rather than at import time.

The inline guard moves into `platform/VoiceSessionSetup.ts` as `requireSetupStrategy()`, which returns the registered strategy or throws. When no strategy is registered *and* the runtime looks like React Native, it names `@elevenlabs/react-native`; otherwise the existing browser message is unchanged, word for word.

Environment detection uses both signals the issue mentions — `navigator.product === "ReactNative"` and the Hermes `HermesInternal` global — read off `globalThis` so neither access throws where the global is absent.

This only changes which *message* an already-throwing call produces. A session that would have started still starts, and a browser app that hits the guard sees exactly what it saw before.

Scoped deliberately to the voice path, where the React Native setup actually matters (WebRTC globals, audio session). `TextConversation` doesn't consult a setup strategy, so nothing there needed to change.

## Verification

Six new tests in the existing `platform/VoiceSessionSetup.test.ts`. `setupStrategy` is module-level state with no reset, so the cases that depend on it re-import the module through `vi.resetModules()` rather than trying to unset it.

Fail-first was done by neutralising `isReactNativeEnvironment()` to `return false` — the pre-fix behaviour, one message for everyone — rather than by stashing the whole module, which would only have produced import errors and proved nothing:

**2 failed / 10 passed**, and the two failures are exactly the React Native cases (`navigator.product` and the Hermes global). The other ten pass in both states, including the two that matter as controls: the browser message is unchanged, and a browser-like environment must *not* mention React Native.

Full `@elevenlabs/client` suite: **206 passed / 16 files**, up from 200 by exactly the six added. `turbo lint` and `check-types` green across all 9 packages.

---

> [!NOTE]
> **Low Risk**
> Only changes error messages when setup was already missing; successful voice sessions and browser guard text are unchanged.
> 
> **Overview**
> When a voice session starts without a registered setup strategy, the guard moves from `VoiceConversation` into **`ensureSetupStrategy()`** in `VoiceSessionSetup.ts`. Behavior is unchanged except for **which error text is thrown**.
> 
> If the runtime looks like React Native (`navigator.product === "ReactNative"` or a `HermesInternal` global), the message tells developers to add `import "@elevenlabs/react-native";` before other ElevenLabs imports instead of pointing at the browser export (which would be wrong on native). Browser and other environments still get the existing browser-entry-point message word for word.
> 
> `VoiceConversation.startSession` now calls `ensureSetupStrategy()(fullOptions)` instead of checking `setupStrategy` inline. Tests cover RN vs browser messaging, Hermes detection, and registered strategies via fresh module imports.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 3af4f1f730797c6bd8132f05326fda2d99bdcfc8. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/packages/client/src/platform/VoiceSessionSetup.ts
+++ b/packages/client/src/platform/VoiceSessionSetup.ts
@@ -56,7 +56,7 @@ function isReactNativeEnvironment(): boolean {
  * registers its own strategy — pointing at the browser entry point there would
  * be actively wrong.
  */
-export function requireSetupStrategy(): VoiceSessionSetupStrategy {
+export function ensureSetupStrategy(): VoiceSessionSetupStrategy {
   if (setupStrategy) {
     return setupStrategy;
   }
--- a/packages/client/src/platform/VoiceSessionSetup.ts
+++ b/packages/client/src/platform/VoiceSessionSetup.ts
@@ -32,6 +32,45 @@ export function setSetupStrategy(strategy: VoiceSessionSetupStrategy) {
 }
 
 /**
+ * Detects a React Native runtime.
+ *
+ * `navigator.product === "ReactNative"` is the long-standing signal; Hermes,
+ * the default engine on modern React Native, exposes a `HermesInternal` global.
+ * Either is enough, and both are absent in a browser.
+ */
+function isReactNativeEnvironment(): boolean {
+  const globals = globalThis as {
+    navigator?: { product?: string };
+    HermesInternal?: unknown;
+  };
+  return (
+    globals.navigator?.product === "ReactNative" ||
+    globals.HermesInternal != null
+  );
+}
+
+/**
+ * Returns the registered setup strategy, or throws explaining how to register
+ * one. On React Native the fix is to import `@elevenlabs/react-native`, which
+ * polyfills the WebRTC globals, configures the native audio session and
+ * registers its own strategy — pointing at the browser entry point there would
+ * be actively wrong.
+ */
+export function requireSetupStrategy(): VoiceSessionSetupStrategy {
+  if (setupStrategy) {
+    return setupStrategy;
+  }
+  throw new Error(
+    isReactNativeEnvironment()
+      ? "No voice session setup strategy registered. It looks like you're using " +
+          "@elevenlabs/client in React Native without importing @elevenlabs/react-native. " +
+          'Add `import "@elevenlabs/react-native";` before any other ElevenLabs imports.'
+      : "No voice session setup strategy registered. " +
+          'Import the platform-specific entry point (e.g. @elevenlabs/client via the "browser" export).'
+  );
+}
+
+/**
  * Sets up a voice session for a WebRTC connection.
  * Platform-agnostic: extracts the input/output controllers that
  * WebRTCConnection already provides via LiveKit.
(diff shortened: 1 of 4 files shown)


packages/client/src/platform/VoiceSessionSetup.ts
```diff
@@ -31,6 +31,45 @@ export function setSetupStrategy(strategy: VoiceSessionSetupStrategy) {
...
+}
+
+/**
+ * Returns the registered setup strategy, or throws explaining how to register
+ * one. On React Native the fix is to import `@elevenlabs/react-native`, which
+ * polyfills the WebRTC globals, configures the native audio session and
+ * registers its own strategy — pointing at the browser entry point there would
+ * be actively wrong.
+ */
+export function requireSetupStrategy(): VoiceSessionSetupStrategy {
```
kraenhansen: I've used the word "ensure" for invariants like these in the past, to not conflate with CJS and `require(...)` statements.

```suggestion
export function ensureSetupStrategy(): VoiceSessionSetupStrategy {
```

kraenhansen (approved):

kraenhansen:

PR #946, opened 2026-08-17
fix(client): key the worklet module cache by the requested source

chinmayv095:
## The problem

`createWorkletModuleLoader` caches the resolved module URL under the worklet's **name** only:

```ts
const cachedUrl = URLCache.get(name);
if (cachedUrl) {
  return worklet.addModule(cachedUrl);
}
```

The `path` argument is never part of the key, so whichever call lands first decides what every later call gets — regardless of what that later call asked for.

The consequence that matters is for `workletPaths`, which exists so an app under a strict CSP can avoid `blob:` and `data:` in `script-src`:

```ts
await load(worklet);                      // no path → caches "blob:…" under "rawAudioProcessor"
await load(worklet, "/vendor/raw.js");    // cache hit → loads the blob, not the path
```

The second call is correctly configured and is silently served the `blob:` URL it was specifically configured to avoid — the CSP-safe path is defeated for the rest of the document's lifetime. `rawAudioProcessor` is loaded from more than one place (`input.ts` on the WebSocket path, `webAudioAdapter.ts` on the WebRTC one), so the ordering that triggers this is a normal sequence of sessions on one page, not a contrived one.

Two self-hosted paths collide the same way: after `load(worklet, "/vendor/a.js")`, a later `load(worklet, "/vendor/b.js")` adds `/vendor/a.js`.

## The fix

Include the requested source in the cache key — the path when one is given, the name alone for the inlined case:

```ts
function cacheKey(name: string, path?: string) {
  return path ? `${name} ${path}` : name;
}
```

Entries stay separate per requested source, and all three write sites store under the same computed key. Repeated identical requests still hit the cache exactly as before; only requests that differ now get their own entry.

## Verification

Eight new tests in a new `createWorkletModuleLoader.test.ts`. The cache is module-level state, so each case re-imports the module via `vi.resetModules()` rather than sharing a cache.

Fail-first was done by reverting `cacheKey` to returning `name` — the current behaviour — giving **3 failed / 5 passed**.

The three failures are exactly the three bug cases (blob-then-path, path-A-then-path-B, path-then-blob). The five passes are controls that must hold in **both** states, and they are the reason to believe this is a narrow fix rather than a cache that has been quietly disabled:

- repeated inlined loads still create exactly **one** blob URL (`URL.createObjectURL` called once across three loads)
- a repeated identical path reuses its entry and never falls through to the blob branch
- separate worklet names keep separate entries
- a failing path still reports the path it tried, and is not cached

`URL.createObjectURL`/`revokeObjectURL` are spied rather than stubbed via `vi.stubGlobal("URL", …)`, which would replace the `URL` constructor and break the module graph.

Full `@elevenlabs/client` suite: **202 passed / 16 files**, up from 194 by exactly the eight added. Prettier, eslint and `tsc --build` clean.

## Note

This is the follow-up I flagged as out of scope in #943, which fixes the *other* half — the WebRTC path not passing `workletPaths` to the loader at all. The two are independent and touch different files; this one stands on its own, and neither depends on the other landing first.

---

> [!NOTE]
> **Low Risk**
> Targeted caching fix in web audio worklet loading with new unit tests; no auth, data, or API surface changes.
> 
> **Overview**
> Fixes **AudioWorklet** loading when apps mix default inlined loads with CSP-friendly **`workletPaths`**: the module-level URL cache no longer keys only on worklet name, so a later self-hosted path is not replaced by an earlier **`blob:`** (or **`data:`**) URL cached under the same name.
> 
> `createWorkletModuleLoader` now uses a **`cacheKey(name, path?)`** (`name` for inlined, `name + path` when a path is passed) for lookups and all cache writes. Repeated identical requests still share one entry; failed path loads are still not cached.
> 
> Adds **`createWorkletModuleLoader.test.ts`** (eight cases, fresh module import per test) and a patch changeset for **`@elevenlabs/client`**.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit 86dbc205e88f51a6fb80cfa032c22ef061ad62d7. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>

### DIFF
--- a/.changeset/worklet-cache-path-key.md
+++ b/.changeset/worklet-cache-path-key.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/client": patch
+---
+
+Key the worklet module cache by the requested source as well as the worklet name, so a self-hosted `workletPaths` entry is no longer served an inlined `blob:` URL cached by an earlier load.
--- a/packages/client/src/platform/web/createWorkletModuleLoader.test.ts
+++ b/packages/client/src/platform/web/createWorkletModuleLoader.test.ts
@@ -1,0 +1,139 @@
+import { afterEach, beforeEach, describe, expect, it, vi } from "vitest";
+
+// The URL cache is module-level state, so every case works against a freshly
+// imported copy of the module rather than a cache left over from the last one.
+async function freshLoader(name = "rawAudioProcessor", source = "// worklet") {
+  vi.resetModules();
+  const { createWorkletModuleLoader } =
+    await import("./createWorkletModuleLoader.js");
+  return createWorkletModuleLoader(name, source);
+}
+
+function createWorklet() {
+  return { addModule: vi.fn().mockResolvedValue(undefined) };
+}
+
+let blobUrlCounter = 0;
+
+// These run in a real browser, so Blob/btoa are genuine. Only the object-URL
+// helpers are replaced, and by spying rather than stubbing the whole `URL`
+// global — overwriting it would take the `URL` constructor with it.
+beforeEach(() => {
+  blobUrlCounter = 0;
+  vi.spyOn(URL, "createObjectURL").mockImplementation(
+    () => `blob:mock-${++blobUrlCounter}`
+  );
+  vi.spyOn(URL, "revokeObjectURL").mockImplementation(() => {});
+});
+
+afterEach(() => {
+  vi.restoreAllMocks();
+});
+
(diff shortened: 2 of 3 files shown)


kraenhansen (approved):

kraenhansen:

Issue #855, opened 2026-06-24
Add MCP tool approval handler option on Conversation


kraenhansen:
## Context

Consumers that want to gate MCP tool calls behind a UI prompt or policy decision today have to wire two things by hand:

1. Subscribe to `onMCPToolCall` and filter for `state === "awaiting_approval"`.
2. Resolve their approval logic, then manually call `sendMCPToolApprovalResult(toolCallId, isApproved)`.

The wiring is mechanical and easy to get wrong (forgetting to ack a `tool_call_id`, double-acking, mishandling state transitions). There's a close precedent in the SDK already: `handleClientToolCall` auto-invokes a user-registered async handler and forwards the result over the wire with try/catch error reporting. Approval flows fit the same shape.

References:

- [`packages/client/src/types.ts#L97`](https://github.com/elevenlabs/packages/blob/main/packages/client/src/types.ts#L97) — current `onMCPToolCall` definition
- [`packages/client/src/BaseConversation.ts#L636-L641`](https://github.com/elevenlabs/packages/blob/main/packages/client/src/BaseConversation.ts#L636-L641) — `sendMCPToolApprovalResult`
- [`packages/client/src/BaseConversation.ts#L275-L332`](https://github.com/elevenlabs/packages/blob/main/packages/client/src/BaseConversation.ts#L275-L332) — `handleClientToolCall` (the pattern to mirror)

## Proposal

Add a new top-level option on `Conversation` alongside `clientTools` (not inside `Callbacks`):

```ts
export type MCPToolApprovalHandler = (
  toolCall: Extract<
    McpToolCallClientEvent["mcp_tool_call"],
    { state: "awaiting_approval" }
  >,
) => Promise<boolean>;

export type MCPToolApprovalConfig = {
  onMCPToolApprovalRequest?: MCPToolApprovalHandler;
};
```

Dispatch in `handleMCPToolCall`, mirroring `handleClientToolCall`:

```ts
protected async handleMCPToolCall(event: MCPToolCallClientEvent) {
  if (this.options.onMCPToolCall) {
    this.options.onMCPToolCall(event.mcp_tool_call);
  }

  const toolCall = event.mcp_tool_call;
  if (
    toolCall.state === "awaiting_approval" &&
    this.options.onMCPToolApprovalRequest
  ) {
    try {
      const approved = await this.options.onMCPToolApprovalRequest(toolCall);
      this.sendMCPToolApprovalResult(toolCall.tool_call_id, approved);
[...]

PR #952, opened 2026-08-18
feat(client): add an MCP tool approval handler option

chinmayv095:
Refs #855 — implements the `@elevenlabs/client` half of that issue (its "1. Vanilla" section). Deliberately **not** using a closing keyword: the issue also proposes a `@elevenlabs/react` sub-provider and hook, which this PR does not add, so it should stay open. See "What is not here" below.

## What this adds

`onMCPToolApprovalRequest`, a top-level option alongside `clientTools`:

```ts
const conversation = await Conversation.startSession({
  agentId: "…",
  onMCPToolApprovalRequest: async toolCall =>
    window.confirm(`Allow ${toolCall.tool_name} from ${toolCall.service_id}?`),
});
```

The SDK sends the `mcp_tool_approval_result` for you, mirroring how `handleClientToolCall` invokes a registered handler and forwards its result. Placement, dispatch site and `Promise<boolean>` return type follow the issue's sketch.

## The part that is more than plumbing

The mechanical version — "await the handler, then ack" — is about ten lines, and it is wrong in the ways @rpelevin set out in the issue thread: on this surface a duplicate, stale or orphaned response is a security bug, not a UI bug. So the handler is backed by a per-`tool_call_id` ledger (`pending` → `settled` | `stale`) which pins the contract:

1. **Exactly one terminal result per `tool_call_id`.** The ledger entry is created before the handler runs and is only ever settled once.
2. **A throw is an explicit deny, not a hang.** Reported through `onError`, then denied.
3. **A late decision is dropped, not sent.** If the call leaves `awaiting_approval` — it ran, failed, or its approval window elapsed — or the session ends while the handler is still out, the entry goes `stale`: the result is discarded and `onError` says so. Without this, a handler resolving after a server-side timeout sends a fresh approval for a call the server has already closed.
4. **A repeat `awaiting_approval` cannot overwrite an unresolved approval.** An id already in the ledger is never dispatched twice, so a retransmission can neither race a pending decision nor contradict one already sent.
5. **The decision binds to the id it was shown for.** The handler receives the narrowed `awaiting_approval` payload, and the ack is keyed off that same object rather than off any later event.
6. **Nothing is dispatched during teardown.** An `awaiting_approval` event can still be delivered between the start of `endSessionWithDetails` and the socket actually closing; the handler is not run for it, so approval UI never appears for a conversation that is already gone.

Two additions beyond a literal reading of the sketch, both deliberate:

- **Non-boolean resolutions fail closed.** `undefined` (or anything else) is reported through `onError` and treated as a denial, so only an explicit `true` can let a tool call through. This also contains a real hazard in the React layer: `mergeOptions` composes same-named functions and discards their return values, so a handler passed both as a provider prop and to `startSession` would otherwise resolve to `undefined` — silently, and on the permissive side if `undefined` were not rejected.
- **An `AbortSignal` in the handler's second argument** (open question 5 in the issue). Without it, a stale approval leaves the consumer's promise pending forever and their modal on screen with no way to know the answer no longer matters. It is additive and ignorable — drop `MCPToolApprovalRequestContext` and the `AbortController` field if you would rather defer it; nothing else in the change depends on it.

## The issue's open questions, and what this takes

- **Naming** — `onMCPToolApprovalRequest`, which is what every example in the issue uses and what matches the `on*` convention. Happy to rename to `mcpToolApprovalHandler`; it is one option key and unreleased.
- **Granularity** — single global handler, per the issue's "single is the simpler start". A consumer can dispatch on `service_id`/`tool_name` inside it.
- **Failure path** — fail closed, per the issue. Not raised upstream: the wire format has no way to distinguish "handler errored" from "denied", and `onError` already carries the distinction locally.
- **SDK-side timeout** — none, per the issue's design decision. Staleness is detected from the server's own subsequent events and from session teardown, so `approval_timeout_secs` needs no client-side timer.
- **Cancellation** — `AbortSignal`, as above.
- **Queueing** — a React-layer question; no opinion taken here. The client is already correct for concurrent approvals (they are independent ledger entries), which is what a queue implementation would need underneath.

## What is not here

The React sub-provider, `useConversationMCPToolApproval`, and the prop-vs-sub-provider precedence rule. Three reasons: open question 4 (the registration channel for a non-`Callbacks` handler) is a design decision that belongs to whoever owns `ConversationContext`; queue-vs-single-slot is unresolved; and #743/#744/#745/#749 are already reworking that exact machinery, so a sub-provider written against today's `registerCallbacks` would conflict with them. This layer is usable from React today by passing the option through `startSession`; adding it to `HookOptions` is a one-line follow-up once the registration question is settled.

Also unchanged: `onMCPToolCall` still fires for every state including `awaiting_approval`, and calling `sendMCPToolApprovalResult()` by hand behaves exactly as before — the ledger only governs calls the handler took ownership of. Existing consumers see no behaviour change.

## Tests

17 new tests in `BaseConversation.test.ts`, including @rpelevin's regression fixture: two approvals outstanding, the second denied first, the first resolving only after its call has moved on — exactly one terminal result is sent, and it carries the right `tool_call_id`.

Fail-first with only `BaseConversation.ts` stashed: **13 failed / 43 passed**. The four new tests that still pass are the controls that must hold in both states — the back-compat case (no handler configured: `onMCPToolCall` fires, nothing is sent) and the three non-`awaiting_approval` states (`loading`, `success`, `failure` must never ask for approval). For the teardown guard specifically, disabling only its condition (rather than the whole module) fails exactly its two tests and leaves the other 54 passing.

With the change: `packages/client` **232 passed / 17 files** (215 before, +17). `@elevenlabs/react` 140 passed / 13 files, unchanged. `turbo lint` and `check-types` green across all 9 packages (29/29 tasks).

Changeset included, `minor` on `@elevenlabs/client`. README documents the option, the fail-closed rule and the abort signal.

---

> [!NOTE]
> **Medium Risk**
> Changes MCP approval wire behavior for opt-in users and introduces fail-closed security rules; mistakes could deny tools or leak stale approvals, though default sessions are unchanged.
> 
> **Overview**
> Adds an opt-in **`onMCPToolApprovalRequest`** session option so apps can approve or deny MCP tools in `awaiting_approval` without calling `sendMCPToolApprovalResult()` themselves. The SDK awaits a boolean, sends **`mcp_tool_approval_result`**, and keeps **`onMCPToolCall`** behavior unchanged for observability; manual approval via `onMCPToolCall` still works when no handler is set.
> 
> Approval handling is backed by a per-`tool_call_id` ledger (**pending → settled | stale**) so each id gets at most one wire result: duplicate `awaiting_approval` events are ignored, late decisions after the call moves on or the session ends are dropped (with **`onError`**), and session teardown aborts in-flight handlers via an **`AbortSignal`** in the handler context. Handler throws and non-boolean resolutions **fail closed** (deny + `onError`). README and a minor changeset document the API; **15** new `BaseConversation` tests cover concurrency, staleness, and back-compat.
> 
> <sup>Reviewed by [Cursor Bugbot](https://cursor.com/bugbot) for commit f36d18453a068d7158345d65f7807082f34d3602. Bugbot is set up for automated code reviews on this repo. Configure [here](https://www.cursor.com/dashboard/bugbot).</sup>


chinmayv095:
Good catch from Bugbot, and it is reachable: an `awaiting_approval` event can still be delivered between the start of `endSessionWithDetails` and the socket actually closing, and the ledger only staled approvals that already existed. A late one opened a fresh pending entry, so approval UI could appear for a conversation that was already gone and the decision would have been settled onto a closing connection.

Fixed in f36d184 by refusing to dispatch when the status is `disconnecting` or `disconnected` — the complement of the guard `endSessionWithDetails` already uses, so the connected path is untouched. Two tests added; disabling only the new condition fails exactly those two.

kraenhansen (approved):

PR #962, opened 2026-08-19
feat(client): support external_agent_disconnected handback

teotoplak:
## Background

When a text conversation is escalated to a human agent, the orchestrator sends `external_agent_connected` and the chat UI enters external-agent mode, where it starts honouring `agent_typing` to show that a *human* is typing on the other end.

There was no way back. When the human closed the ticket, the backend teardown ran but nothing reached the UI: the chat stayed stuck in human mode indefinitely, with the typing indicator potentially stuck on.

[elevenlabs/xi#42700](https://github.com/elevenlabs/xi/pull/42700) adds the server half, emitting a new `external_agent_disconnected` event on handback and firing an AI turn so the agent announces it has resumed control. **This PR is the client half.**

It exists because of [this review comment](https://github.com/elevenlabs/xi/pull/42700#discussion_r3797579314): the xi PR only updated the two dashboard chat UIs, which parse WebSocket events by hand. There are three independent chat UIs, and the public embeddable widget consumes this repo instead.

| UI | Where | Covered by |
|---|---|---|
| Dashboard agent preview | xi, hand-rolled WS types | xi#42700 |
| Dashboard conversation view | xi, hand-rolled WS types | xi#42700 |
| **Public embeddable widget** | **this repo** | **this PR** |

(`services/js/convai-widget` in xi never implemented external-agent mode, so it is not a fourth gap.)

## What changed

- **`@elevenlabs/types`** - mirror the event into `agent.asyncapi.yaml` (message, payload, channel list, event enum) and regenerate. The generated files are pure generator output; I re-ran `generate-types` and the result was byte-identical to the committed diff.
- **`@elevenlabs/client`** - new `onExternalAgentDisconnected` callback, handler, switch case, and event union member, mirroring the existing `external_agent_connected` path at every site.
- **`@elevenlabs/react`** - re-export the callback through `HookCallbacks`.
- **`@elevenlabs/convai-widget-core`** - leave external-agent mode and clear the typing indicator on handback.

## Testing

Created playwright test (widget test), also did manual inspection - typing indicators stop (not a lot to showcase).

### DIFF
--- a/.changeset/light-crews-repeat.md
+++ b/.changeset/light-crews-repeat.md
@@ -1,0 +1,8 @@
+---
+"@elevenlabs/types": minor
+"@elevenlabs/client": minor
+"@elevenlabs/react": minor
+"@elevenlabs/convai-widget-core": minor
+---
+
+Add support for the `external_agent_disconnected` client event. The client SDK exposes a new `onExternalAgentDisconnected` callback, and the widget leaves external-agent mode (and clears the typing indicator) when the external human agent disconnects and the AI agent resumes control.
--- a/packages/client/src/BaseConversation.ts
+++ b/packages/client/src/BaseConversation.ts
@@ -16,6 +16,7 @@ import type {
   AgentTypingEvent,
   ClientToolCallEvent,
   ExternalAgentConnectedEvent,
+  ExternalAgentDisconnectedEvent,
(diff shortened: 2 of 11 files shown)


kamilk-11 (approved):
LGTM

PR #966, opened 2026-08-20
[Agents] Hide language selector on collapsed widget trigger

kalvin807:
[AWF-89](https://linear.app/elevenlabs/issue/AWF-89/language-dropdown-visibility-feedback)
[Slack](https://eleven-labs-workspace.slack.com/archives/C08A0QJ8YQH/p1780661426842489)

**Why:** Since v0.13.0 the language dropdown sits on the collapsed launcher. Customers who sized an iframe to that launcher (HCLTech) get overflow that blocks clicks on the page behind it. They want the picker back in the expanded sheet only.

**What:**
- Add `show_language_selector_on_trigger` (embed: `show-language-selector-on-trigger`). Default `true` keeps current launcher behavior.
- `false` hides the picker on collapsed expandable and non-expandable triggers. The expanded sheet header still shows it under existing sheet rules.
CleanShot 2026-08-20 at 16 06 17

**Test:** `ExpandableTriggerActions.test.ts` — launcher hides when `false`; sheet still shows the picker after expand.

### DIFF
--- a/.changeset/show-language-selector-on-trigger.md
+++ b/.changeset/show-language-selector-on-trigger.md
@@ -1,0 +1,6 @@
+---
+"@elevenlabs/convai-widget-core": minor
+"@elevenlabs/convai-widget-embed": minor
+---
+
+Add a `show_language_selector_on_trigger` widget config option (and matching `show-language-selector-on-trigger` embed attribute). Default `true` keeps the language dropdown on the collapsed launcher. Set to `false` to hide it there; the expanded sheet header still shows it.
--- a/.changeset/language-selector-placement.md
+++ b/.changeset/language-selector-placement.md
@@ -1,6 +1,0 @@
----
-"@elevenlabs/convai-widget-core": minor
-"@elevenlabs/convai-widget-embed": minor
----
-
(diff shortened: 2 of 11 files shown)


giannagerton (approved):

.changeset/show-language-selector-on-trigger.md
```diff
@@ -0,0 +1,6 @@
+---
+"@elevenlabs/convai-widget-core": minor
+"@elevenlabs/convai-widget-embed": minor
+---
+
+Add a `show_language_selector_on_trigger` widget config option (and matching `show-language-selector-on-trigger` embed attribute). Default `true` keeps the language dropdown on the collapsed launcher. Set to `false` to hide it there; the expanded sheet header still shows it.
```
kamilk-11, line 6: This option only affects the visibility of the dropdown, but `v0.13.0` also changed the overall behavior, right? I.e. the primary button that would previously only expand the widget, now immediately starts the call. Just want to mention it here as I'm not sure if the customer is interested in just removing the dropdown or brining back the previous experience completely

kamilk-11 (approved):

.changeset/show-language-selector-on-trigger.md
kalvin807: Thanks for the flagging it. Will introduce 1 more flag to revert that if they follow up.

PR #967, opened 2026-08-20
feat(widget): render the concurrency wait-queue status

raghavan-gv:
## What

Widget follow-up to elevenlabs/xi#46137: the embeddable ConvAI widget now renders the concurrency wait-queue lifecycle from the `queue_status` server events (`waiting` / `admitted` / `timed_out`).

- No `@elevenlabs/client` changes: the SDK forwards unhandled server events to `onDebug` on both the WebSocket and WebRTC transports. Once the queue is finalized then we will make the SDK changes.

## Testing

Tested by embedding widget in a local server connecting it to local backend 

https://github.com/user-attachments/assets/7c1cd93a-9215-44d3-95d8-b2d75ed3325e

### DIFF
--- a/packages/convai-widget-core/src/contexts/conversation.tsx
+++ b/packages/convai-widget-core/src/contexts/conversation.tsx
@@ -575,6 +575,8 @@ function useConversationSetup() {
         conversationRef.current?.sendFeedback(like);
       },
       sendUserMessage: (text: string, options?: SendUserMessageOptions) => {
+        // The orchestrator discards messages sent while held in the queue.
+        if (isWaitingForAgent.peek()) return;
         conversationRef.current?.sendUserMessage(text, options);
         transcript.value = [
           ...transcript.value,
@@ -591,6 +593,7 @@ function useConversationSetup() {
         text?: string;
         file: TranscriptFileInput & { fileId: string };
       }) => {
+        if (isWaitingForAgent.peek()) return;
         const trimmed = input.text?.trim() ?? "";
         const { fileId, ...fileInput } = input.file;
         conversationRef.current?.sendMultimodalMessage({
--- a/packages/convai-widget-core/src/contexts/conversation.tsx
+++ b/packages/convai-widget-core/src/contexts/conversation.tsx
@@ -88,6 +88,10 @@ export type TranscriptEntry =
       type: "mode_toggle";
       mode: ConversationMode;
       conversationIndex: number;
+    }
+  | {
+      type: "queue_timeout";
+      conversationIndex: number;
     };
 
 export function ConversationProvider({ children }: ConversationProviderProps) {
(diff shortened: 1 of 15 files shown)


packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -404,7 +427,27 @@ function useConversationSetup() {
             onExternalAgentConnected: () => {
               isExternalAgentMode.value = true;
             },
+            // The SDK forwards server events it does not handle (such as
+            // queue_status) to onDebug, alongside other debug payloads, so
+            // the event shape has to be narrowed here.
```
kamilk-11: Are there plans to make the SDK handle those explicitly?

packages/convai-widget-core/src/widget/AvatarOverlay.tsx
```diff
@@ -57,7 +57,10 @@ export function AvatarOverlay({
           </div>
         </InOutTransition>
         <InOutTransition active={showStatusLabel}>
-          <div className="absolute -bottom-3 left-1/2 -translate-x-1/2 translate-y-full transition-[opacity,transform] data-hidden:opacity-0 data-hidden:scale-75">
+          {/* w-max: an absolutely positioned box at left-1/2 otherwise
+              shrink-to-fits against half the avatar container, collapsing
+              wrapped copy (the queue waiting status) to min-content width. */}
```
kamilk-11: I wrote a statement on ai generated comments like this one some time ago: https://eleven-labs-workspace.slack.com/archives/C096NU6LJBX/p1784110668716609
This PR hits all three points I mentioned

packages/convai-widget-core/src/widget/StatusLabel.tsx
```diff
@@ -7,38 +7,82 @@ import { useTextContents } from "../contexts/text-contents";
...
   ...props
-}: HTMLAttributes<HTMLDivElement>) {
-  const currentLabel = userCurrentLabel();
-  const [label, setLabel] = useState(currentLabel.peek().label);
+}: StatusLabelProps) {
+  const currentLabel = userCurrentLabel(compact);
+  const [{ label, wrap }, setLabel] = useState(() => {
+    const { label, wrap } = currentLabel.peek();
+    return { label, wrap };
+  });
```
kamilk-11: Is the double destructuring needed here?
```suggestion
  const [{ label, wrap }, setLabel] = useState(() => currentLabel.peek());
```

packages/convai-widget-core/src/widget/StatusLabel.tsx
```diff
@@ -53,7 +97,15 @@ export function StatusLabel({
...
     >
       <InOutTransition key={label} initial={false} active={true}>
-        <div className="animate-text whitespace-nowrap transition-[opacity,transform] ease-out duration-200 data-hidden:opacity-0 transform data-hidden:translate-y-2">
+        <div
+          className={clsx(
+            "animate-text transition-[opacity,transform] ease-out duration-200 data-hidden:opacity-0 transform data-hidden:translate-y-2",
+            // truncate degrades gracefully when a surface caps the label's
+            // width (e.g. long custom copy in the trigger): ellipsis instead
+            // of spilling past the card edge.
+            wrap ? "whitespace-normal max-w-60 text-center" : "truncate"
```
kamilk-11: But the `wrap` is hardcoded per label type, so the customer specifying a longer custom copy for the speaking status, for example, will not have the wrapping applied here. Shouldn't wrapping be enabled always when compact is false?

kamilk-11 (commented):
LGTM

packages/convai-widget-core/src/contexts/conversation.tsx
raghavan-gv: Yes, once the queue protocol is finalized we'll add a class onQueueStatus to the SDK and we can drop this later.

packages/convai-widget-core/src/widget/AvatarOverlay.tsx
raghavan-gv: Makes total sense, I've removed and shortened most of the unnecessary comments.

packages/convai-widget-core/src/widget/StatusLabel.tsx
raghavan-gv: Done, thanks.

packages/convai-widget-core/src/widget/StatusLabel.tsx
raghavan-gv: Right, wrap should be a property of the surface rather than label. I've removed wrap entirely and the layout is derived from compact.

kamilk-11 (approved):

PR #970, opened 2026-08-20
[Widget] Show first message in text mode for Text & Voice agents

giannagerton:
## Summary

Fixes [ACF-55](https://linear.app/elevenlabs/issue/ACF-55) 
[Slack thread](https://eleven-labs-workspace.slack.com/archives/C08A0QJ8YQH/p1775855454543779) 

`conversation.tsx` drops the first message for any text conversation that has a `first_message` configured, since it arrives after the user's opening message and would render out of order. `Sheet` compensated by rendering it locally — but only for agents configured text-only, so Text & Voice agents got neither copy. 

`Sheet` now renders the local copy for any agent that supports text mode and has the text input enabled: as a preview before connecting, and through a text conversation. Voice conversations are untouched.

Behavior (text input turned on, transcript off)

 **Before** no first message when opening the widget via the chat button
before-first-message

**After** first message shows, ability to switch to voice
after

### DIFF
--- a/packages/convai-widget-core/src/widget/Sheet.tsx
+++ b/packages/convai-widget-core/src/widget/Sheet.tsx
@@ -2,6 +2,7 @@ import { useComputed, useSignal } from "@preact/signals";
 import {
   useFirstMessage,
   useIsConversationTextOnly,
+  useTextInputEnabled,
   useTextOnly,
   useWidgetConfig,
 } from "../contexts/widget-config";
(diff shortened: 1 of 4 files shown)


aarkalyk (approved):

packages/convai-widget-core/src/widget/Sheet.tsx
```diff
@@ -50,22 +52,44 @@ export function Sheet({ open }: SheetProps) {
...
+
+    if (isConversationTextOnly.value) return message;
+
+    const showFirstMessage =
+      isDisconnected.value &&
+      config.value.supports_text_only &&
+      textInputEnabled.value &&
+      !transcript.value.some(
+        entry => entry.type === "message" && !entry.isText
+      );
```
kamilk-11: This would read easier imo due to fewer negations:
```suggestion
      transcript.value.every(
        entry => entry.type !== "message" || entry.isText
      );
```

kamilk-11 (approved):
Nice!

PR #980, opened 2026-08-25
feat(widget): send multiple files per multimodal message

bsreenivas11:
## Summary
- `useFileUpload` holds `PendingFile[]` and the attach input is `multiple`.
- Composer sends `fileIds` and renders N pending chips / transcript attachments.
- Depends on the client `fileIds` PR (`bharath/multi-file-upload-client`).

Written by Cursor Grok 4.6, using Cursor

## Test plan
- [ ] Attach one file — still works
- [ ] Attach several files, send, confirm chips and transcript
- [ ] Disable attach at conversation `maxFiles`
- [ ] Merge after the client PR

### DIFF
--- a/packages/convai-widget-core/src/contexts/conversation.tsx
+++ b/packages/convai-widget-core/src/contexts/conversation.tsx
@@ -50,6 +50,7 @@ export type TranscriptEntry =
       conversationIndex: number;
       eventId?: number;
       fileInput?: TranscriptFileInput | null;
+      fileInputs?: TranscriptFileInput[] | null;
     }
   | {
       type: "agent_tool_request";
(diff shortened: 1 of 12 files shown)


packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -50,6 +50,7 @@ export type TranscriptEntry =
       conversationIndex: number;
       eventId?: number;
       fileInput?: TranscriptFileInput | null;
+      fileInputs?: TranscriptFileInput[] | null;
```
giannagerton, line 53: We should be able to remove `fileInput` and just have `fileInputs` here since this type isn't exported from the package.

packages/convai-widget-core/src/utils/display-transcript.ts
```diff
@@ -24,6 +24,7 @@ export type DisplayTranscriptEntry =
       eventId?: number;
       toolStatus?: ToolCallStatusType;
       fileInput?: TranscriptFileInput | null;
+      fileInputs?: TranscriptFileInput[] | null;
```
giannagerton, line 27: Same here, let's see if we can remove `fileInput` in favor of just `fileInputs`

giannagerton (approved):
LGTM just a few comments to look at before merging

PR #981, opened 2026-08-25
feat(client): add fileIds to sendMultimodalMessage

bsreenivas11:
## Summary
- Add `fileIds` to `sendMultimodalMessage` on `@elevenlabs/client` (plural wins over deprecated `fileId`).
- Dual-send `file` + `files` so older backends still receive the first attachment.
- Changeset for `@elevenlabs/client`, `@elevenlabs/react`, and `@elevenlabs/types`.

Widget follow-up: `bharath/multi-file-upload-widget`.

Written by Cursor Grok 4.6, using Cursor

## Test plan
- [x] `BaseConversation` tests for singular, plural, prefer-plural, and text-only
- [ ] CI on this PR
- [ ] Publish `@elevenlabs/react` after merge so xi can pin `fileIds`

### DIFF
--- a/.changeset/multi-file-upload-client.md
+++ b/.changeset/multi-file-upload-client.md
@@ -1,0 +1,7 @@
+---
+"@elevenlabs/client": minor
+"@elevenlabs/react": minor
+"@elevenlabs/types": patch
+---
+
+Add `fileIds` to `sendMultimodalMessage` and dual-send `file` + `files` on the wire.
--- a/packages/client/src/BaseConversation.test.ts
+++ b/packages/client/src/BaseConversation.test.ts
@@ -417,6 +417,80 @@ describe("BaseConversation", () => {
     });
   });
 
+  describe("sendMultimodalMessage", () => {
+    function conversationSending() {
+      const sendMessage = vi.fn();
+      const connection = {
+        ...noopConnection,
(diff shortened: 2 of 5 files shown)


.changeset/multi-file-upload-client.md
```diff
@@ -0,0 +1,7 @@
+---
+"@elevenlabs/client": minor
+"@elevenlabs/react": minor
+"@elevenlabs/types": patch
```
AngeloGiacco, line 4: how come some minor and some patch?

AngeloGiacco (approved):

.changeset/multi-file-upload-client.md
bsreenivas11: client minor — new public option fileIds on sendMultimodalMessage. That is a feature.

react minor — so XI can bump @elevenlabs/react and get fileIds through the re-exported MultimodalMessageInput. This PR does not change any react source.

types patch — only an optional files? on the generated multimodal_message type. Existing callers still typecheck, so we treated it as a compatible type tweak.

AngeloGiacco: hm ok fair enough was just kinda thinking id expect them to all be the same but your reasoning makes sense

PR #992, opened 2026-09-01
[Widget] First message rich content language

giannagerton:
### Read first message rich content from the active language preset

Adds language support for first message rich content. `first_message_rich_content` is now read from the active language preset with a fallback to the top level value. Switching the language in the widget before a conversation starts updates the quick reply buttons text to the translated value.


kamilk-11 (approved):
LGTM

PR #995, opened 2026-09-02
[Widget] fix first agent message getting dropped

giannagerton:
### Summary
Fix the first agent reply being dropped in text-only widget conversations when a configured first message is interrupted.

### Root cause
[xi PR #47525](https://github.com/elevenlabs/xi/pull/47525) made backend first-message generation asynchronous. Although the widget had already displayed the greeting locally, its server-side turn could still be running when the user replied. If interrupted, its final agent_response never arrived, causing the widget to mistake and suppress the agent’s actual reply.

This exposed a latent widget bug from [packages PR #345](https://github.com/elevenlabs/packages/pull/345): the widget continued suppressing agent streams until that final greeting response arrived. When it never arrived, the real reply was suppressed too.

Before

https://github.com/user-attachments/assets/0ed330c9-2473-41bd-b0e1-c23019c91e92

After

https://github.com/user-attachments/assets/a56fab30-1bba-48bc-8fd1-97e63e334d50

### Changes
- Release the first-message guard when the greeting stream begins.
- Continue suppressing the configured greeting’s final event (event_id: 1) when it arrives.
- Regression tests

### DIFF
--- a/.changeset/fix-text-chat-first-agent-reply.md
+++ b/.changeset/fix-text-chat-first-agent-reply.md
@@ -1,0 +1,5 @@
+---
+"@elevenlabs/convai-widget-core": patch
+---
+
+Fix the agent's first reply being dropped from the transcript in text chat when the agent has a first message configured.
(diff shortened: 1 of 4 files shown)


kamilk-11 (approved):
Thanks!

PR #996, opened 2026-09-03
[Widget] Fix duplicate widget messages before tool calls

giannagerton:
Fixes AWF-296 [AWF-296: Agent's final message not rendered in widget for text-only conversations](https://linear.app/elevenlabs/issue/AWF-296/agents-final-message-not-rendered-in-widget-for-text-only)
Fixes AWF-297 [AWF-297: Widget renders an agent message twice when the message is immediately followed by a tool call](https://linear.app/elevenlabs/issue/AWF-297/widget-renders-an-agent-message-twice-when-the-message-is-immediately)

### Summary
Fixes two widget transcript issues caused by agent streaming events arriving in different orders around tool calls:

- An agent message could render twice immediately before a tool call.
- A final post-tool agent message could disappear before disconnection.

### Root cause
The widget tracked streaming with a global index and boolean that were reset independently.

After a stream stopped, the index could be cleared while the receiving flag remained active. A subsequent `agent_response` could overwrite a newer stream placeholder producing a duplicate or find no stream index and return without rendering, dropping the message.

### Fix
Replace the separate streaming refs with one consolidated stream state that tracks:

- pending stream slots awaiting canonical responses
- the currently active stream
- a complete response awaiting its streamed copy

Final responses now:
1. Match pending streams by `event_id`.
2. Prefer an exact-text match for late responses.
3. Consolidate into the transcript tail when safe.
4. Append normally when no safe match exists.
Immediate streamed copies of an already rendered response are ignored. Stream state is reset between sessions and on disconnect.

**Message duplicated before tool call**

Before

https://github.com/user-attachments/assets/67e1451d-8ad6-4200-92bf-c31483b02c26

After

https://github.com/user-attachments/assets/e7d0338e-7006-44e1-b7bb-e65abb679755

**Final message no longer gets dropped**

Before

https://github.com/user-attachments/assets/3d2c5165-b565-4a4a-b666-d0e30984eed7

After

https://github.com/user-attachments/assets/5026895c-e8bd-49b7-97ce-5a7240d6325c

### DIFF
--- a/packages/convai-widget-core/src/contexts/conversation.tsx
+++ b/packages/convai-widget-core/src/contexts/conversation.tsx
@@ -31,6 +31,52 @@ import { useShadowHost } from "./shadow-host";
 
 const FIRST_MESSAGE_EVENT_ID = 1;
 
+type AgentEventId = number | undefined;
+type AgentMessageKey = string;
+
+type AgentMessagePointer = {
+  index: number;
+  eventId: AgentEventId;
+};
+
+type AgentStream = AgentMessagePointer & {
+  kind: "stream";
+  key: AgentMessageKey;
+};
+
+type IgnoredAgentStream = {
+  kind: "ignored";
+  key: AgentMessageKey;
+  eventId: AgentEventId;
+};
+
+type AgentStreamState = {
+  pending: Map<AgentMessageKey, AgentStream>;
+  active: AgentStream | IgnoredAgentStream | null;
+  unmatchedResponses: Map<AgentMessageKey, AgentMessagePointer>;
+  streamOccurrences: Map<AgentEventId, number>;
+  responseOccurrences: Map<AgentEventId, number>;
+};
+
+function createAgentStreamState(): AgentStreamState {
+  return {
+    pending: new Map(),
+    active: null,
+    unmatchedResponses: new Map(),
+    streamOccurrences: new Map(),
+    responseOccurrences: new Map(),
+  };
+}
+
+function nextAgentMessageKey(
+  occurrences: Map<AgentEventId, number>,
+  eventId: AgentEventId
+): AgentMessageKey {
+  const occurrence = occurrences.get(eventId) ?? 0;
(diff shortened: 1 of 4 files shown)


packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -363,6 +430,12 @@ function useConversationSetup() {
                   eventId: event_id,
                 },
               ];
+              if (role === "agent") {
+                agentStreamStateRef.current.unmatchedResponse = {
```
kamilk-11: I think it could be:
```suggestion
                streamState.unmatchedResponse = {
```

packages/convai-widget-core/src/contexts/conversation.tsx
```diff
@@ -378,9 +451,35 @@ function useConversationSetup() {
...
-                streamingMessageIndexRef.current = currentTranscript.length;
+                const streamState = agentStreamStateRef.current;
+                const unmatchedResponse = streamState.unmatchedResponse;
+                const unmatchedEntry =
+                  unmatchedResponse == null
+                    ? undefined
+                    : currentTranscript[unmatchedResponse.index];
+                if (
+                  unmatchedResponse?.index === currentTranscript.length - 1 &&
+                  unmatchedResponse.eventId === event_id &&
```
kamilk-11, line 483: There's an annoying issue with `eventId` which is that it's only unique per turn, not per message, and a single turn can have multiple agent messages. For example, when an agent decides to transfer to another agent, it may do:
1. generate text response talking about the transfer
2. generate a tool call to perform the transfer
3. transferred agent generates its greeting message

All of the above would have the same `event_id` unfortunately and currently the BE does not send any other differentiator, so to be completely correct we'd need to count event ids for `agent_response` and each type of `agent_chat_response_part`, and use that to derive the unique id, i.e. `${eventId}-${numberOfTimesThisEventIDOccuredForThisSpecificEvent}`

kamilk-11 (approved):
LGTM, Left a comment

PR #1007, opened 2026-09-09
[Widget] Fix duplicate reply after tool calls

giannagerton:
Fixes [AWF-305: Procedure call causes duplicate messages](https://linear.app/elevenlabs/issue/AWF-305/procedure-call-causes-duplicate-messages)
### The problem
The agent message directly after a tool call renders twice in the widget. This reproduces with procedures, which call tools as part of their steps. This is a regression from #996
### Root cause
#996 paired each stream segment with `agent_response` by arrival order within an `event_id`, which assumes every segment gets a response. A tool call breaks that: the turn opens a stream segment that carries no text and no response, then streams the reply in a later segment of the same turn. The counters end up off by one, so the reply's canonical response is written into the earlier, stale segment while the streamed copy stays where it was rendered.
### Fix
Match a canonical response to a pending stream slot by transcript contentsinstead of arrival order. A response claims, in preference order: the slot already showing this exact text, the first still-empty slot, then the oldest slot of the turn.
### Testing
Added four new mock agents to cover the tool call shapes


kamilk-11 (approved):
LGTM

PR #1012, opened 2026-09-11
fix(connection): debounce user_activity

mat-sz:
Linear issue: [AWF-299](https://linear.app/elevenlabs/issue/AWF-299/add-debounce-to-senduseractivity-in-sdk)

Added 2s debounce to `sendUserActivity` to avoid spamming the backend (the timing is in line with how backend debounces the events).

### DIFF
--- a/packages/client/src/BaseConversation.ts
+++ b/packages/client/src/BaseConversation.ts
@@ -7,7 +7,7 @@ import type {
   FormatConfig,
 } from "./utils/BaseConnection.js";
 import { uploadFile, type UploadFileResult } from "./utils/uploadFile.js";
-import { debounce, type DebouncedFunction } from "./utils/debounce.js";
+import { throttle, type ThrottledFunction } from "./utils/throttle.js";
 import type { Conversation } from "./index.js";
 import type {
   AgentAudioEvent,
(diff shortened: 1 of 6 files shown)


packages/client/src/utils/debounce.ts
```diff
@@ -0,0 +1,66 @@
...
+
+/**
+ * Creates a debounced wrapper around `fn` that limits how often it runs within
+ * a `waitMs` window.
+ *
+ * With the default (`leading: true`, `trailing: false`) configuration the first
+ * call invokes `fn` immediately and any further calls within `waitMs` are
+ * suppressed until the window elapses.
+ */
+export function debounce<Args extends unknown[]>(
```
kamilk-11: Nit but this is more of a throttle than debounce

packages/client/src/BaseConversation.ts
```diff
@@ -44,6 +45,8 @@ const END_CALL_DETAILS: DisconnectionDetails = {
   context: { type: "end_call", reason: "Agent ended the call" },
 };
 
+const USER_ACTIVITY_DEBOUNCE_MS = 2000;
```
kamilk-11: Our USER_ACTIVITY_THRESHOLD_SECS is exactly 2 seconds so this is really on the edge, I'd go for something like 1 - 1.5

kamilk-11 (approved):

packages/client/src/utils/debounce.ts
mat-sz: Renamed.

packages/client/src/BaseConversation.ts
mat-sz: Updated to `1000`.

PR #1023, opened 2026-09-21
feat(scribe): add transcriptEdit option and edited_transcript event

JacekDuszenko:
Adds `transcriptEdit` to `Scribe.connect()` and dispatches the new `edited_transcript` server message, following the pattern of #932 (entity detection).

- `transcriptEdit?: string`: natural-language instruction applied to each committed transcript (max 2000 characters), sent as the `transcript_edit` query param. Cannot be combined with `entityDetection`; the server rejects the combination with `invalid_request`.
- New server message `edited_transcript` → `RealtimeEvents.EDITED_TRANSCRIPT` / `EditedTranscriptMessage` with `text` (the committed text the instruction was applied to) and `edited_text` (equal to `text` when nothing changed). Edits run asynchronously and may arrive out of commit order, so `text` is the field to correlate on.
- `scribe.asyncapi.yaml` updated and `@elevenlabs/types` regenerated (additive only: `EditedTranscript`, `EditedTranscriptMessage`).
- Tests: URI includes `transcript_edit` when set and omits it otherwise; `edited_transcript` is dispatched to its listener.
- `@elevenlabs/react`: `useScribe()` accepts `transcriptEdit`, exposes an `onEditedTranscript` callback, and attaches the edited text to the matching `committedTranscripts` entry as `editedText` (matched on the committed `text`, since edits may arrive out of commit order).
- Changeset: `@elevenlabs/client`, `@elevenlabs/types` and `@elevenlabs/react` minor.

Upstream: the parameter and event ship in elevenlabs/xi#49618 (websocket route) and elevenlabs/xi#51266 (public AsyncAPI spec); docs in elevenlabs/elevenlabs-dx#2778. Pricing: 30% premium on the base transcription cost, billed for at least 10 seconds of audio per committed transcript.

Verified locally: repo-wide `pnpm run lint` (pre-commit) green; client `vitest run --browser.headless src/scribe/scribe.test.ts` 62 passed; react `vitest run --browser.headless src/scribe.test.tsx` 8 passed.


PaulAsjes (commented):
You're missing the updates to the React package: https://github.com/elevenlabs/packages/blob/main/packages/react/src/scribe.ts

PaulAsjes (approved):

